PR #354, opened 2022-03-02
[FIX] Fix device_strs

zhuohan123:
@zhisbug Is this the right fix?


merrymercy:
@zhisbug Should we merge this?

zhisbug:
> @zhisbug Should we merge this?

I'll do an experiment on this one today.

PR #360, opened 2022-03-14
Add unittest for resharding

ZYHowell:


alpa/testing.py
```diff
@@ -201,6 +213,117 @@ def loss_func(params):
         return train_step
 
 
+def test_resharding(var,
```
merrymercy, line 216: Move this to `alpa/tests/test_cross_mesh_resharding.py`

alpa/testing.py
```diff
@@ -370,3 +493,49 @@ def run_n_layer_bert(self,
         hlo_text = executable.get_hlo_text()
         executable.shutdown()
         return hlo_text
+
+
+class ReshardingTest(unittest.TestCase):
```
merrymercy, line 498: Move this to `alpa/tests/test_cross_mesh_resharding.py`

merrymercy (changes requested):

PR #367, opened 2022-03-25
[CI/CD] build jax, jaxlib, alpa, and test

zhisbug:


merrymercy:
Installing alpa is very easy. The tricky part is to install jaxlib. Can we wrap jax, jaxlib, and alpa all into a single wheel?

zhisbug:
I planed to release a jax-alpa wheel to pypi, and set it as a dependency of alpa

PR #368, opened 2022-03-29
[CI/CD] Initiate a build docker

zhisbug:


.gitignore
```diff
@@ -1,7 +1,6 @@
 # Python cache
 __pycache__
 *.pyc
-build
```
ZYHowell, line 4: xla customcall marker has a build folder, but should be ignored

zhisbug: yeah I fixed it in #367

PR #369, opened 2022-04-01
[Documentation] install flax==0.4.0

PKUFlyingPig:
The installation guide `pip3 install flax` will install flax 0.4.1 in default, which is not compatible with current jax-alpa (jax==0.2.26). The resulting error may occur when import alpa, which is quite obscure.


zhisbug:
LGTM thanks for the contributions.

PR #388, opened 2022-04-14
Add fwd-only support to pipeshard

yf225:


tests/test_auto_sharding_fwd_only.py
```diff
@@ -0,0 +1,77 @@
...
+
+        def func(x):
+            bufs_updated = {}
+            bufs_updated["_dummy"] = jnp.ones(1)
+            x = jnp.matmul(x, x)
+            x = jnp.matmul(x, x)
+            y = x + 1
+            return bufs_updated, (x, y)
+
+        def wrapper_func(x):
```
yf225, line 43: This is very much a hack because I wasn't able to get fwd-only to work without having a apply_grad stage (XLA would just return empty protos). So here I add a dummy apply_grad stage that satisfies these properties: 1. all outputs are tracked, 2. the dummy operations in the apply_grad stage are simple so that it doesn't take up much compute.

yf225:
Alpa team will provide better approach. Closing this.

PR #390, opened 2022-04-14
[RUNTIME] Save/Load DistributedArray on each Host node distributedly

PKUFlyingPig:
I added two methods for class DistributedArray: 
- save(path)  will save only one replica of the data that the DistributedArray's remote_buffers point to. This function can be called in the driver and it will dispatch the saving process to the MeshHostWorker, which uses TensorStore package to save the DeviceArray into EFS.
- load(path) will dispatch the loading process to the MeshHostWorker, which uses TensorStore package to load the data stored in EFS into the buffers.
The saved and loaded DistributedArray can have different ShardingSpec, so I think this can be the starting point for the model's distributed save/load.

You can also check the test_dist_save_load.py unit test.

### DIFF
--- a/alpa/device_mesh.py
+++ b/alpa/device_mesh.py
@@ -7,6 +7,7 @@ from itertools import chain
 import logging
 from operator import attrgetter
 import os
+import re
 import time
 from typing import Any, List, Union, Sequence, Tuple, Optional
 
@@ -28,6 +29,7 @@ from jax.lib import xla_client
 import jax.numpy as jnp
 import numpy as np
 import ray
+import tensorstore as ts
 
 from alpa import mesh_profiling
 import alpa.collective as col
@@ -96,6 +98,72 @@ class MeshHostWorker:
                     jax_tensor_to_cupy(device_put(
                         jnp.ones((1,), dtype=jnp.int8), d),
                                        take_ownership=True))
+
+    ##### TensorStore Related Functions #####
+    def get_tensorstore_spec(self, ckpt_path: str):
+        spec = {
+            'driver': 'zarr',
+            'kvstore': {},
+            'metadata_key': f".zarray{self.host_id}"
+        }
+        if ckpt_path.startswith('gs://'):
+            m = re.fullmatch('^gs://([^/]*)/(.*)$', ckpt_path, re.DOTALL)
+            if m is None:
+                raise ValueError(
+                    'The ckpt_path should contain the bucket name and the '
+                    f'file path inside the bucket. Got: {ckpt_path}')
+            gcs_bucket = m.group(1)
+            path_without_bucket = m.group(2)
+            spec['kvstore'] = {
+                'driver': 'gcs',
+                'bucket': gcs_bucket,
+                'path': path_without_bucket
+            }
+        else:
+            spec['kvstore'] = {'driver': 'file', 'path': ckpt_path}
+        return spec
+
+    def load_buffers_from_ts_store(self, ckpt_dir: str, uuids: Sequence[int],
+                     shard_indices: Sequence[Index], device_ids: Sequence[int]):
+        ts_spec = self.get_tensorstore_spec(ckpt_dir)
+        t = ts.open(ts.Spec(ts_spec), open=True).result()
+
+        for index, uuid, device_id in zip(shard_indices, uuids, device_ids):
+            data = t[index].read().result()
+            self.put_buffer(uuid, device_id, data)
+
+    def save_buffers_to_ts_store(self, ckpt_dir: str, uuids: Sequence[int],
+                     shard_indices: Sequence[Index],
+                     global_shape: Sequence[int]):
+        for uuid in uuids:
+            assert uuid in self.buffers
+
+        ts_spec = self.get_tensorstore_spec(ckpt_dir)
+        dtype = self.buffers[uuids[0]].dtype
+        if dtype == jnp.bfloat16:
+            # Tensorstore uses 'bfloat16', not '<V2'.
+            dtype = 'bfloat16'
+        else:
+            dtype = np.dtype(dtype).str
+        metadata = {
+            'compressor': {
+                'id': 'gzip'
+            },
+            'shape': global_shape,
+            'chunks': self.buffers[uuids[0]].shape,
+            'dtype': dtype,
+        }
+        ts_spec['metadata'] = metadata
+        t = ts.open(ts.Spec(ts_spec),
+                    create=True,
+                    open=True,
+                    context=ts.Context({'file_io_concurrency': {
+                        'limit': 128
+                    }})).result()
+
+        for index, uuid in zip(shard_indices, uuids):
+            t[index].write(self.buffers[uuid]).result()
+
 
     ##### Buffer Related Functions #####
     def put_buffer(self, uuid: int, device_id: int, data: np.ndarray):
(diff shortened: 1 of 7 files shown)


alpa/device_mesh.py
```diff
@@ -3,12 +3,16 @@
...
+from importlib.metadata import metadata
 from itertools import chain
 import logging
 from operator import attrgetter
 import os
 import time
 from typing import Any, List, Union, Sequence, Tuple, Optional
+import pathlib
+import re
+import tensorstore as ts
```
zhuohan123, line 15: Can you add tensorstore to the dependency list [here](https://github.com/alpa-projects/alpa/blob/main/docs/install/from_source.rst)?

alpa/device_mesh.py
```diff
@@ -97,9 +101,86 @@ def __init__(self, server_address: str, num_hosts: int, host_id: int):
...
+            self.put_buffer(uuid, device_id, data)
+        
+        for index, uuid, device_id in zip(shard_indices, uuids, device_ids):
+            _load_array(index, uuid, device_id)
+ 
+
+
     def put_buffer(self, uuid: int, device_id: int, data: np.ndarray):
-        assert uuid not in self.buffers
+        #assert uuid not in self.buffers
```
zhuohan123, line 183: Why comment out this line?

alpa/device_mesh.py
```diff
@@ -1123,6 +1217,61 @@ def delete(self):
     def flush(self):
         self._npy_value = None
 
+    #======== distributed save/load ========== START#
```
zhuohan123, line 1220: Change this comment to follow the existing style like the following:
```
##### distributed save/load #####
```

tests/test_dist_save_load.py
```diff
@@ -0,0 +1,169 @@
...
+class DeviceMeshTest(unittest.TestCase):
+
+    def setUp(self):
+        os.environ["XLA_PYTHON_CLIENT_ALLOCATOR"] = "platform"
+        ray.init(address="auto", ignore_reinit_error=True)
+
+    def tearDown(self):
+        ray.shutdown()
+    
+    def test_distributed_array_save_load(self):
```
zhuohan123, line 30: This test only works for an 8-GPU cluster. Change this to work for more general clusters or add an if to skip this test for clusters of other shapes.

tests/test_dist_save_load.py
```diff
@@ -0,0 +1,169 @@
...
+
+        # Check results
+        assert_allclose(optimizer_expected.target, optimizer_actual.target)
+
+        physical_mesh.shutdown()
+
+def suite():
+    suite = unittest.TestSuite()
+    suite.addTest(DeviceMeshTest("test_distributed_array_save_load"))
+    # suite.addTest(DeviceMeshTest("test_mlp"))
```
zhuohan123, line 162: What's the status of this test?

zhuohan123 (changes requested):
Also, we do have a [formatter](https://alpa-projects.github.io/developer/developer_guide.html#formatting-and-linting). Please format the code before merging it into the main branch.

alpa/device_mesh.py
```diff
@@ -97,9 +101,86 @@ def __init__(self, server_address: str, num_hosts: int, host_id: int):
...
+            gcs_bucket = m.group(1)
+            path_without_bucket = m.group(2)
+            spec['kvstore'] = {'driver': 'gcs', 'bucket': gcs_bucket,
+                            'path': path_without_bucket}
+        else:
+            spec['kvstore'] = {'driver': 'file', 'path': ckpt_path}
+        return spec
+
     ##### Buffer Related Functions #####
+    def view_buffer(self, uuid: int):
```
merrymercy, line 121: This function is redundant.
Use `get_buffer` instead

alpa/device_mesh.py
```diff
@@ -97,9 +101,86 @@ def __init__(self, server_address: str, num_hosts: int, host_id: int):
...
+        ts_spec['metadata'] = metadata
+        t = ts.open(
+            ts.Spec(ts_spec),
+            create=True,
+            open=True,
+            context=ts.Context({'file_io_concurrency': {
+                'limit': 128
+            }})).result()
+        
+        def _write_array(index, uuid):
```
merrymercy, line 157: code style: why do not you inline this function?

tests/test_dist_save_load.py
```diff
@@ -0,0 +1,169 @@
...
+                                                    (global_input_data1,))
+
+        # Check the DistributedArray's remote buffers
+        self.assertEqual(
+        ['[[0]\n [2]]', '[[1]\n [3]]', '[[4]\n [6]]', '[[5]\n [7]]', 
+        '[[ 8]\n [10]]', '[[ 9]\n [11]]', '[[12]\n [14]]', '[[13]\n [15]]'],
+        dist_input_data1.dist_view())
+
+        # Save the DistributedArray (one replica only)
+        save_path = "/home/ubuntu/efs/save_test/"
```
merrymercy, line 62: Can we avoid this hard-coded path?

alpa/device_mesh.py
```diff
@@ -97,9 +101,86 @@ def __init__(self, server_address: str, num_hosts: int, host_id: int):
...
+        else:
+            spec['kvstore'] = {'driver': 'file', 'path': ckpt_path}
+        return spec
+
     ##### Buffer Related Functions #####
+    def view_buffer(self, uuid: int):
+        assert uuid in self.buffers
+        return str(self.buffers[uuid])
+
+    def save_buffers(self,
```
merrymercy, line 125:
```suggestion
    def save_buffers_to_ts_store(self, 
```
Also, move this function under `TensorStore Related Functions`

alpa/device_mesh.py
```diff
@@ -1123,6 +1217,61 @@ def delete(self):
     def flush(self):
         self._npy_value = None
 
+    #======== distributed save/load ========== START#
+    
+    def dist_view(self):
```
merrymercy, line 1222: Use `self.device_mesh.get_remote_buffers(self.remote_buffers, batching=True)` and compare the numerical values in unit test instead of comparing strings

alpa/device_mesh.py
```diff
@@ -1123,6 +1217,61 @@ def delete(self):
...
+        """Return a string representation of the remote buffers"""
+        obj_refs = []
+        for buf_ref in self.remote_buffers:
+            obj_refs.append(self.device_mesh.workers[buf_ref.host_id].view_buffer.remote(buf_ref.uuid))
+        return ray.get(obj_refs)
+
+
+    def save(self, path: str):
+        """Save one replica of the DeviceArrays that self.remote_buffers point to distributedly."""
```
merrymercy, line 1231:
```suggestion
        """Save one replica of the array to `path` distributedly."""
```
Do not add unnecessary low-level details into the docstring.

alpa/device_mesh.py
```diff
@@ -1123,6 +1217,61 @@ def delete(self):
...
+                self.device_mesh.workers[host_id].save_buffers.remote(
+                                                        path, 
+                                                        uuids, 
+                                                        indices_per_host[host_id],
+                                                        self.shape)
+                            )
+        return ray.get(obj_refs)
+    
+    def load(self, path: str):
+        """Load the data in **path** into the DeviceArrays that self.remote_buffers point to distributedly."""
```
merrymercy, line 1251:
```suggestion
        """Load the data from `path` distributedly."""
```

tests/test_dist_save_load.py
```diff
@@ -0,0 +1,169 @@
...
+        input_sharding_spec = logical_mesh.make_tile_spec(global_input_data1, [0, 1], [0, 1])
+        input_indices = input_sharding_spec.indices(global_input_data1.shape).flatten()
+        (dist_input_data1,) = physical_mesh.shard_args_to_arrays(
+                                                    (jax.ShapedArray(global_input_data1.shape, jnp.int32),),
+                                                    (input_indices,),
+                                                    (input_sharding_spec,),
+                                                    (global_input_data1,))
+
+        # Check the DistributedArray's remote buffers
+        self.assertEqual(
```
merrymercy, line 56: Why do you compare strings instead of numerical values?

merrymercy (changes requested):

alpa/device_mesh.py
PKUFlyingPig: This function is debug-only, I will remove it after I implement model save/load. Leave this as resolved to remind me.

PKUFlyingPig:
I changed the load(path) into a classmethod for DistributedArray, which serves as a factory method to load the data from tensorstore, put onto device and return a DistributedArray object. Currently it needs four arguments: path, aval, device_mesh and sharding_spec. 
However, this brought a problem when I tried to implement the model save/load: When I want to load a model, how can I get the sharding_specs for each tensor in the model. The intuitive way may use alpa.parallel to shard the model first, but this requires the tensor has already been loaded into the model. 
A simple workaround approach: When the user wants to load a model, he initializes a new model, parallelize it and then replace all the remote buffers with previously saved params. This approach is simple to implement, but it loses the essential benefit of "distributed loading" that the model size may overflow the driver's device memory. 
Also, since Alpa's current API simply decorates the training_loop function, which shards the model after the model initialization, do we assume the model size will not overflow the driver's device memory?

PKUFlyingPig:
Suggested by Zhuohan, it is possible to get the parallelized executable with an abstract input model, which contains the device_mesh and sharding_spec for each tensor in the model. I will follow this approach to implement model loading.

alpa/device_mesh.py
merrymercy: Is this comment resolved?

tests/test_dist_save_load.py
```diff
@@ -0,0 +1,151 @@
...
+        load_sharding_spec = logical_mesh.make_tile_spec(
+            global_input_data1, [0, 1], [0])
+        dist_load_data1 = DistributedArray.load(tmpdir.name, jax.ShapedArray(global_input_data1.shape, jnp.int32), physical_mesh, load_sharding_spec)
+
+        # Check the DistributedArray's remote buffers
+        remote_buffers2 = dist_load_data1.device_mesh.get_remote_buffers(dist_load_data1.remote_buffers, batching=True)
+        desired_buffers2 = [[[0, 1], [2, 3]], [[0, 1], [2, 3]], [[4, 5], [6, 7]],[[4, 5], [6, 7]], 
+                            [[8, 9], [10, 11]], [[8, 9], [10, 11]], [[12, 13], [14, 15]], [[12, 13], [14, 15]]]
+
+        for b1, b2 in zip(desired_buffers2, remote_buffers2):
```
merrymercy, line 82: This can be done in one line. Try something like
```assert_allclose(np.array(remote_buffers2), np.array(desired_buffers2))```

merrymercy (changes requested):
The updated version looks good to me.
We can merge this PR first and leave the model save/load to the next PR. To do so, delete the whole `DistSaveLoadTest::test_mlp` function instead of commenting it out.

PKUFlyingPig:
I will work on the model save/load on another branch in my fork and initiate another PR after I implement it.

merrymercy (commented):
The git history is messed up in https://github.com/alpa-projects/alpa/pull/390/files which includes changes from other pull requests.
You can rebase to main instead of merging to clean the history.

PKUFlyingPig:
I thought I already used `git pull --rebase` to rebase to main?

PKUFlyingPig:
Sorry for the messed-up git history, I manually changed all the modifications into one single commit. Hope this can help.

zhuohan123 (commented):
LGTM. Please format your changes before you merge.

merrymercy (approved):

zhuohan123 (approved):
LGTM

PR #430, opened 2022-05-07
[FIX] fix bug in stage construction without apply grad

ZYHowell:


zhuohan123 (approved):
LGTM

merrymercy:
Why do you merge it without passing CI?

PR #436, opened 2022-05-08
[RUNTIME] Accelerate sharding arrays

merrymercy:
- Use numpy indexing operations to shard XLA CPU device arrays. When used together with ray, the new method is 3x - 10x faster than the old jax.jit implementation.
- misc updates on the micro-benchmark scripts.


ZYHowell:
The new impl of shard device array is faster because the jax device array is on CPU, but with LocalPhysicalDeviceMesh, the device array is on the accelerator, so there is a memcpy. Currently I use the local one on TPU, but I'll try to support DistributedPhysicalDeviceMesh.

PR #437, opened 2022-05-08
[RUNTIME] Distributed Save and Load for Model

PKUFlyingPig:
I added two top-level APIs for Alpa: 
- save_checkpoint(ckpt_dir, target, step): the `target` is usually the trainState. This function (1) replaces the leaves of trainState (DistributedArray or ReplicatedDistributedArray) with the path to which the array is saved and saves this metadata for restore. (2) saves one replica for all the DistributedArrays/ReplicatedDistributedArrays in `target`.
- restore_checkpoint(executable, ckpt_dir, target, step): This function requires the DecentralizedRuntime which contains all the shardingSpec and mesh allocation info to load and reshard the DistributedArrays/ReplicatedDistributedArrays. 

Current version contains a defect that the trainState must be the first argument in the parallelized function (i.e. train_step). One possible fix requires the user to explicitly provide the index of the trainState in train_step's argument list. Any suggestion for a more elegant implementation?

### DIFF
--- a/alpa/__init__.py
+++ b/alpa/__init__.py
@@ -3,7 +3,7 @@ from alpa.data_loader import DataLoader, MeshDriverDataLoader
 from alpa.device_mesh import (DeviceCluster, PhysicalDeviceMesh,
                               LocalPhysicalDeviceMesh,
                               DistributedPhysicalDeviceMesh, DistributedArray,
-                              ReplicatedDistributedArray, fetch)
+                              fetch)
 from alpa.global_env import global_config, set_parallelize_options
 from alpa.mesh_profiling import ProfilingResultDatabase
 from alpa.pipeline_parallel.primitive_def import (mark_pipeline,
--- a/alpa/__init__.py
+++ b/alpa/__init__.py
@@ -2,8 +2,8 @@ from alpa.api import clear_callable_cache, grad, parallelize, value_and_grad
 from alpa.data_loader import DataLoader, MeshDriverDataLoader
 from alpa.device_mesh import (DeviceCluster, PhysicalDeviceMesh,
                               LocalPhysicalDeviceMesh,
-                              DistributedPhysicalDeviceMesh, DistributedArray, ReplicatedDistributedArray,
-                              fetch)
+                              DistributedPhysicalDeviceMesh, DistributedArray,
+                              ReplicatedDistributedArray, fetch)
 from alpa.global_env import global_config, set_parallelize_options
 from alpa.mesh_profiling import ProfilingResultDatabase
 from alpa.pipeline_parallel.primitive_def import (mark_pipeline,
--- a/alpa/__init__.py
+++ b/alpa/__init__.py
@@ -2,7 +2,7 @@ from alpa.api import clear_callable_cache, grad, parallelize, value_and_grad
 from alpa.data_loader import DataLoader, MeshDriverDataLoader
 from alpa.device_mesh import (DeviceCluster, PhysicalDeviceMesh,
                               LocalPhysicalDeviceMesh,
-                              DistributedPhysicalDeviceMesh, DistributedArray,
+                              DistributedPhysicalDeviceMesh, DistributedArray, ReplicatedDistributedArray,
                               fetch)
 from alpa.global_env import global_config, set_parallelize_options
 from alpa.mesh_profiling import ProfilingResultDatabase
@@ -11,6 +11,7 @@ from alpa.pipeline_parallel.primitive_def import (mark_pipeline,
 from alpa.pipeline_parallel.layer_construction import (
     manual_remat, automatic_remat, automatic_layer_construction,
     manual_layer_construction)
+from alpa.serialization import save_checkpoint, restore_checkpoint
 from alpa.shard_parallel.auto_sharding import LogicalDeviceMesh
 from alpa.util import XlaPassContext
 from alpa.timer import timers
--- a/alpa/pipeline_parallel/decentralized_distributed_runtime.py
+++ b/alpa/pipeline_parallel/decentralized_distributed_runtime.py
@@ -204,10 +204,17 @@ class DecentralizedDistributedRuntime(BaseDistributedRuntime):
                          num_batch=num_batch)
         self.uuid_counter = 0  # counter for local buffer uuid
         self.flop_count = flop_count
+        self.in_tree = in_tree
 
         # List[stage_idx -> executable_uuid]
         self.executable_uuids = []
 
+        # Cached sharding indices for inputs.
+        # List[mesh_idx -> List[sharding_indices]].
+        self.input_indices = [] 
+        # Cached sharding specs for inputs.
+        # List[mesh_idx -> List[sharding_spec]]
+        self.input_specs = []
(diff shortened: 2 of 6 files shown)


alpa/__init__.py
```diff
@@ -3,14 +3,15 @@
 from alpa.device_mesh import (DeviceCluster, PhysicalDeviceMesh,
                               LocalPhysicalDeviceMesh,
                               DistributedPhysicalDeviceMesh, DistributedArray,
-                              fetch)
+                              ReplicatedDistributedArray, fetch)
```
zhisbug, line 6: do you use `ReplicatedDistributedArray` somewhere?

PKUFlyingPig:
The root cause of the problem can be understood better with an example:

user's view: train_step(batch, train_state) where batch = {"x": x, "y": y}, train_state = {"a": a,  "params":{"w": w, "b": b}}, 
after tree_flatten: args = [x, y, a, w, b]
after split (batchnum = 2) : args = [x0, x1, y0, y1, a, w, b]

The shardingSpec and device mesh allocation information for all the arguments is stored according to the flatten and split version above. So in order to load train_state, I have to know the index of the first arg of train_state (a in this case) in the flatten and split argument arrays. 

After the discussion with @zhuohan123, we have two possible fixes:
(1) requires the user to explicitly provide the index of the trainState in train_step's argument list.
For the example above, the user provides index 1 in `restore_checkpoint`.
(2) @zhuohan123 argues that the user does not have to provide the entire `DecentralizedDistributedRuntime` into restore_checkpoint but only provide the shardingSpec and device mesh allocation info. To support this, we can add another API in `DecentralizedDistributedRuntime` to get the info.
For the example above, the user calls executable.get_load_info() to get the info list:
[{"x": x_info, "y": y_info}, {"a": a_info, "params": {"w": w_info, "b": b_info}}]
Then users can index into the list and get just enough info for loading and provide it into restore_checkpoint.

alpa/__init__.py
PKUFlyingPig: I use it in the `serialization.py`.

merrymercy:
This problem is very important. Let use move our discussion to https://github.com/alpa-projects/alpa/issues/439

alpa/serialization.py
```diff
@@ -0,0 +1,149 @@
...
+import enum
+import os
+from typing import Union, Any
+import uuid
+
+from flax.serialization import to_state_dict, from_state_dict
+import jax
+import msgpack
+import numpy as np
+from tensorflow.io import gfile
```
zhuohan123, line 15: Why do we need this dependency?

alpa/serialization.py
```diff
@@ -0,0 +1,149 @@
...
+class _MsgpackExtType(enum.IntEnum):
+    """Messagepack custom type ids."""
+    ndarray = 1
+    native_complex = 2
+    npscalar = 3
+    distarray = 4
+    replicated_distarray = 5
+
+
+def _ndarray_to_bytes(arr) -> bytes:
```
zhuohan123, line 32: Can we just import this function from flax instead of copying it?

tests/test_dist_save_load.py
```diff
@@ -73,20 +89,141 @@ def test_distributed_array_save_load(self):
...
+
+        # Run before save
+        serial_state = state
+        parallel_state = state
+        serial_state = serial_train_step(serial_state, batch)[0]
+        parallel_state = parallel_train_step(parallel_state, batch)[0]
+        assert_allclose(serial_state.params, parallel_state.params, 1e-3, 1e-3)
+
+        with tempfile.TemporaryDirectory(
+                prefix="/home/ubuntu/efs/") as ckpt_dir:
```
zhuohan123, line 134: We should not include any absolute path. Can we somehow use ray to hack a temporary NFS just for testing purposes?

zhuohan123 (changes requested):

tests/test_dist_save_load.py
zhuohan123: Or you can force all the testing processes on a same node and just use the filesystem on the node.

alpa/serialization.py
PKUFlyingPig: flax's serialization function uses these packages, maybe I can find some in-built utils to substitute?

zhuohan123: Where exactly does flax use this? I can't find it on this page:https://flax.readthedocs.io/en/latest/_modules/flax/serialization.html

PKUFlyingPig: https://flax.readthedocs.io/en/latest/_modules/flax/training/checkpoints.html#save_checkpoint

zhisbug:
How's this issue going? will it get in main soon?

zhuohan123:
From what I know this PR should be ready to merge now. @PKUFlyingPig Anything else you wanna do before we do a final round of review and merge the code?

PKUFlyingPig:
> From what I know this PR should be ready to merge now. @PKUFlyingPig Anything else you wanna do before we do a final round of review and merge the code?

This code requires tensorflow as a third-party package (flax's serialization needs this), should we mention it in the installation document and add it into the CI/DI scripts. This code failed the CI/Unittest due to this.

alpa/serialization.py
zhuohan123: Let's leave this dependency here for now. However, we should remove the dependency to tensorflow just for this small part of function.

alpa/serialization.py
```diff
@@ -0,0 +1,166 @@
...
+        https://flax.readthedocs.io/en/latest/_modules/flax/training/checkpoints.html#restore_checkpoint
+
+        Args:
+            ckpt_dir: directory of checkpoints to restore from.
+            step: step number to load.
+            target: matching object to rebuild via deserialized state-dict.
+            load_info: shardingSpec and deviceMesh allocation info for loading.
+    """
+    ckpt_path = os.path.join(ckpt_dir, f"checkpoint_{step}")
+    with gfile.GFile(ckpt_path, 'rb') as fp:
```
zhuohan123, line 148: Actually, if the only dependency to gfile is here, can we try to replace this gfile with python's native file and see whether the code still work for our cases?

zhuohan123 (approved):

alpa/serialization.py
PKUFlyingPig: Sure, I will try.

alpa/pipeline_parallel/decentralized_distributed_runtime.py
```diff
@@ -200,10 +203,17 @@ def __init__(self,
...
         self.uuid_counter = 0  # counter for local buffer uuid
         self.flop_count = flop_count
+        self.in_tree = in_tree
 
         # List[stage_idx -> executable_uuid]
         self.executable_uuids = []
 
+        # Cached sharding indices for inputs.
+        # List[mesh_idx -> List[sharding_indices]].
+        self.input_indices = []
```
merrymercy, line 213: This var is deprecated

merrymercy (commented):
Let us remove the tensorflow dependency. Other things look good to me

merrymercy (approved):

PR #447, opened 2022-05-12
[FIX] only requires coin cbc solver

merrymercy:


zhisbug:
was glpk-utils never used? could you update here: https://github.com/alpa-projects/alpa/blob/main/setup.py#L85

merrymercy:
pulp is still required for coin cbc

PR #448, opened 2022-05-12
[FEATURE] Support outvars concatenated in batch dimension

ZYHowell:
Currently we compile a concat executable and it does not support a following apply grad stage. I'll work on it later, but this version is enough for inference.

### DIFF
--- a/alpa/api.py
+++ b/alpa/api.py
@@ -31,7 +31,7 @@ def parallelize(fun=None,
                 static_argnums="auto",
                 donate_argnums="auto",
                 batch_argnums=(1,),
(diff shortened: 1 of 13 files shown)


merrymercy:
Every new functionality should come with a test case.
You should upload your local test cases or modify this https://github.com/alpa-projects/alpa/blob/main/tests/test_pipeline_inference_only.py

merrymercy:
@ZYHowell Every new functionality should come with a test case.
You should upload your local test cases or modify this https://github.com/alpa-projects/alpa/blob/main/tests/test_pipeline_inference_only.py

PR #453, opened 2022-05-15
Allow using CUDA_HOME setting in xla_custom_call_marker build

yf225:
This PR allows user to specify their CUDA install location for xla_custom_call_marker build.


zhuohan123 (approved):
LGTM

Issue #439, opened 2022-05-09
[RFC] A Data Structure for Whole Parallelization Plan


merrymercy:
# Background
Alpa does not have a single data structure to store the whole parallelization plan for a model. The parallelization plan is stored messily across multiple places. Currently, the most similar one is this part of code.
https://github.com/alpa-projects/alpa/blob/3416c2c0ced6f222425867952f355fd514ac1207/benchmark/alpa/suite_auto_gpt.py#L58-L61
This part stores the pipeline stage construction results (i.e., layer-to-stage mapping and mesh shapes) got by the previous search, which can be loaded for future use. However, these informations are not organized in a clean data structure. In addition, some key components such as sharding specs of weight tensors are missing.

# A Data Structure for Whole Parallelization Plan
The major use cases of this data structure include model initialization, model save/load, and strategy serialization. To serve these cases, we need to store the pipeline strategy for the whole model and the sharding strategy for each pipeline stage. This RFC proposes a draft design.

```python
class ParallelPlan:
    pipeline_plan: PipelinePlan
    stage_plans: Sequence[StagePlan]
    weight_placement: Dict[str, TensorPlacement]
    input_placement: Dict[str, TensorPlacement]  # used for the data loader

class PipelinePlan:
    forward_stage_layer_ids
    sub_physical_mesh_shapes
    sub_logical_mesh_shapes
    submesh_autosharding_option_dicts
    layer_construction_options
    micro_batch_size

class StagePlan:
    auto_sharding_solution_vector

class TensorPlacement:
    mesh_ids: Sequence[int]  # A tensor can be replicated on multiple stages
    sharding_specs: Sequence[ShardingSpec]
```

# API design
This data structure can be used for the following cases.

## Model Initialization
Use abstract values to get a parallel plan and initialize models according to the plan. Note that this breaks our one-line parallelization API.
```python
state_aval = return_aval(create_train_state)()
batch_aval = create_batch_aval()
plan = train_step.get_parallel_plan(state_aval, batch_aval)

state = alpa.parallelize(create_train_state, method=plan)()
for batch in data_loader:
    state = train_step(state, batch)
```
Question: when and how to launch physical device meshes.

## Strategy Serialization
We can save the parallelization plans to a file for future usages. The file stores a  
`List[Tuple(train_step_hash, cluster_info, parallel_plan, peak_memory, latency)]`, where `train_step_hash = hash(train_step.jaxpr)`.  
[...]

PR #454, opened 2022-05-15
Add support for in-place initialization of weights on worker

yf225:
This is probably a short-term approach.

### DIFF
--- a/alpa/device_mesh.py
+++ b/alpa/device_mesh.py
@@ -8,7 +8,7 @@ from operator import attrgetter
 import os
 import re
 import time
-from typing import Any, List, Union, Sequence, Tuple, Optional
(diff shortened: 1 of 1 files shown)


alpa/device_mesh.py
```diff
@@ -1122,7 +1152,7 @@ def get_remote_buffers(self,
 
     def delete_remote_buffers(self, buf_refs: List["RemoteBufferRef"]):
         """Delete remote buffers."""
-        if self.workers is None or not ray.is_initialized():
+        if self.workers is None or ray.is_initialized is None or not ray.is_initialized():
```
yf225, line 1155: Without this, it complains `ray` is None during `delete_remote_buffers` phase.

merrymercy:
Yes. This is probably a short-term plan. I will work on a more principal approach following https://github.com/alpa-projects/alpa/issues/439

PR #458, opened 2022-05-18
[TEST] Add more inference test

ZYHowell:
Support returning `last_hidden_state`, `hidden_state` and `attention` in inference.

### DIFF
--- a/tests/test_pipeline_inference_only.py
+++ b/tests/test_pipeline_inference_only.py
@@ -4,16 +4,18 @@ import jax
 import jax.numpy as jnp
 
 from alpa import init, PipeshardParallel
-from alpa.testing import (MLPModel, create_train_state, get_mlp_inference_step,
-                          assert_allclose)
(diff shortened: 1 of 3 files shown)


tests/test_pipeline_inference_only.py
```diff
@@ -4,16 +4,18 @@
...
+from alpa.model.bert_model import BertConfig, FlaxBertLayerCollection
+from alpa.testing import (MLPModel, create_train_state,
+                          get_bert_layer_collection_inference_step,
+                          get_mlp_inference_step, assert_allclose)
 
 
 class PipelineInferenceTest(unittest.TestCase):
+
     def setUp(self):
         init(cluster="ray")
```
merrymercy, line 16: If we do not call `alpa.shutdown()`, the second test will follow the physical mesh group created by the first test. Is this intended?

ZYHowell: it's a mistake

merrymercy (approved):

PR #463, opened 2022-05-23
[RUNTIME] Simplify restore_checkpoint() API

PKUFlyingPig:
The `target` argument in restore_checkpoint is redundant. Use `load_info` instead.


zhuohan123 (approved):
LGTM

PR #468, opened 2022-05-27
[FORMAT] Update formatter and linter from the ray and sky project

zhuohan123:


.github/workflows/ci.yml
```diff
@@ -12,31 +12,8 @@ on:
     branches: [main]
 
 jobs:
-
```
zhisbug, line 15: change to this linter instead of removing linting?

requirements-dev.txt
```diff
@@ -0,0 +1,5 @@
+# formatting
+yapf==0.32.0
```
zhisbug, line 2: remove this file, add it to setup.py https://github.com/alpa-projects/alpa/blob/main/setup.py#L89

.github/workflows/ci.yml
zhuohan123: Linter is now in a separate github action yaml. We still have the linter in `yapf.yml` and `pylint.yml`. Do you think I need to move this part back into `ci.yml`?

zhisbug: can you merge yapf and pylint into one job in ci.yml and let test depend on it (like I did). Otherwise, test will be trigger even if linting is not passed, which wates CI resources

zhuohan123: OK sounds good.

zhisbug:
this linter looks simpler.

- How will it check docstrings (missing and formats)? You can see here propspector applies pep257 after running pylint for docstrings.

- once you are sure it works as expected, you can remove this file: https://github.com/alpa-projects/alpa/blob/main/.prospector.yaml

zhuohan123:
> * How will it check docstrings (missing and formats)? You can see here propspector applies pep257 after running pylint for docstrings.

The new linter does check missing docstrings. But I don't see it check the docstring formats for now. Does prospector check for formats? How strict will it be?

PR #470, opened 2022-05-31
[CI/CD] fix CD bug, update docs with new installation procedures

zhisbug:


docs/install.rst
```diff
@@ -4,9 +4,61 @@ Install Alpa
...
+- CUDA (CuDNN): 11.1 (8.0.5), 11.2 (8.1.0), 11.3 (8.2.0)
+- Python: 3.7, 3.8, 3.9
+
+If you need to use other CUDA, CuDNN, or Python versions, please follow the next section to :ref:`install from source<install-from-source>`.
+
+1. To install from wheels, first install Alpa:
+
+  .. code:: bash
+
+    pip3 install --upgrade pip
```
merrymercy, line 25: Why is this line required?

docs/install.rst
```diff
@@ -80,3 +132,47 @@ You can check the installation by running the following test script.
...
+  >>> import alpa
+    ......
+    RuntimeError: jaxlib version 0.3.7 is newer than and incompatible with jax version 0.3.5. Please update your jax and/or jaxlib packages
+
+Make sure your jax version is 0.3.5, Flax version is 0.4.1 by reinstalling them following:
+
+.. code:: bash
+
+  pip3 install jax==0.3.5
+  pip3 install flax==0.4.1
```
merrymercy, line 176: Does this mean we have to update this webpage every time we release the package?

docs/install.rst
zhisbug: mimicking https://github.com/google/jax#pip-installation-cpu

docs/install.rst
zhisbug: no, why you think so?

PR #473, opened 2022-06-02
[REFACTOR] Add runtime emitter for pipeshard parallel

ZYHowell:
This PR does https://github.com/alpa-projects/alpa/pull/451#pullrequestreview-974178288. After this PR:
- RuntimeEmitter generates a pipeshard config without any call of mesh host worker's methods, so physical meshes can be instantiated later. 
- Resources' instantiation(nccl groups, nccl communicators and worker executables) are in Pipeshard Executable's initialization.

### DIFF
--- a/alpa/device_mesh.py
+++ b/alpa/device_mesh.py
@@ -1771,6 +1771,15 @@ class PhysicalDeviceMeshGroup:
         self.collective_groups[src_mesh_id][dst_mesh_id] = cg
         self.collective_groups[dst_mesh_id][src_mesh_id] = cg
 
+    def instantiate_nccl_group(self,
+                               src_mesh_id: int,
+                               dst_mesh_id: int):
+        cg = self.collective_groups[src_mesh_id][dst_mesh_id]
+        if global_config.eagerly_create_communicators:
+            cg.instantiate_now()
+        else:
+            cg.instantiate()
+
     def instantiate_all(self):
         for src_id in range(len(self)):
             for dst_id in range(src_id + 1, len(self)):
--- a/alpa/device_mesh.py
+++ b/alpa/device_mesh.py
@@ -1747,7 +1747,10 @@ class PhysicalDeviceMeshGroup:
     def index(self, *args, **kwargs):
         return self.meshes.index(*args, **kwargs)
 
-    def establish_nccl_group(self, src_mesh_id: int, dst_mesh_id: int):
+    def establish_nccl_group(self,
+                             src_mesh_id: int,
+                             dst_mesh_id: int,
+                             instantiate=True):
         """Establish NCCL group between two meshes."""
         # pylint: disable=import-outside-toplevel
         from alpa.pipeline_parallel.cross_mesh_resharding import CollectiveGroup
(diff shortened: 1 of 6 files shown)


alpa/device_mesh.py
```diff
@@ -1780,12 +1783,32 @@ def establish_nccl_group(self, src_mesh_id: int, dst_mesh_id: int):
...
+
+    def instantiate_all(self):
+        for src_id in range(len(self)):
+            for dst_id in range(src_id + 1, len(self)):
+                cg = self.collective_groups[src_id][dst_id]
+                if cg and not cg.instantiated:
+                    if global_config.eagerly_create_communicators:
+                        cg.instantiate_now()
+                    else:
+                        cg.instantiate()
```
merrymercy, line 1811: This pattern repeats three times. Can we avoid it?

alpa/pipeline_parallel/compile_executable.py
```diff
@@ -179,9 +180,16 @@ def compile_pipeshard_executable(fun: lu.WrappedFun,
         schedule=schedule,
         is_batch=batch_invars,
         num_batch=num_microbatch,
-        flop_count=total_flops,
-        concat_vars_mapping=concat_vars_mapping,
-        in_tree=in_tree)
+        in_tree=in_tree).compile(concat_vars_mapping)
```
merrymercy, line 183: Why is `concat_vars_mapping` an argument of `compile` instead of an argument of `__init__`?

alpa/pipeline_parallel/pipeshard_executable.py
```diff
@@ -1311,7 +371,7 @@ def __init__(self, worker: MeshHostWorker, uuid: int,
         self.global_buffers = worker.buffers
         self.acc_grad_buffers = {}
         self.acc_in_uuids = [list(uuids) for uuids in list(acc_local_uuids)]
-        self.acc_out_uuids = acc_out_uuids
+        self.acc_out_uuids = [list(uuids) for uuids in list(acc_out_uuids)]
```
merrymercy, line 374: You often convert a numpy array to a list by calling `list()`. Most of them are not necessary and I deleted a lot of this pattern last time. Try to remove them and work directly with numpy indexing.

merrymercy (changes requested):

alpa/device_mesh.py
ZYHowell: Now all of them call the same internal method

alpa/pipeline_parallel/pipeshard_executable.py
ZYHowell: Currently in PipeshardExecutable there is almost no `list(XXX)`. I'll check the PipelineInstEmitter.

alpa/pipeline_parallel/compile_executable.py
ZYHowell: You're right. I move it to init now

alpa/global_env.py
```diff
@@ -20,10 +20,21 @@ def __init__(self):
         self.shard_parallel_sync_for_timer = False
 
         ########## Options of pipeline_parallel ##########
-        self.profile_with_whole_ray_cluster = True
+        # Whether to debug with local runtime. The local runtime checks
+        # correctness of stage construction and other graph level operations.
         self.debug_with_local_runtime = False
```
merrymercy, line 25: Is `debug_with_local_runtime` used? What's the difference between `debug_with_local_runtime` and `debug_with_pipeshard_runtime`

ZYHowell: With `debug_with_pipeshard_runtime`, we can dump everything(`ReshardingTask`, `PipelineInst`, `ExecutableConfig` etc.) without physical resources. `LocalRuntime` aims to check the correctness of passes on jaxprs.

merrymercy: I think `debug_with_local_runtime` is never used in the current codebase

merrymercy (approved):
The refactor looks good to me. Sorry I merged my PR #477 first, which introduces some conflicts.

alpa/global_env.py
ZYHowell: I'll deprecate it later

zhisbug: Is the local runtime still there? I think we can deprecate all those code

Issue #416, opened 2022-04-30
[RUNTIME] NCCL broadcast based cross mesh resharding


TarzanZhao:
**Background**

To send a distributed tensor from one mesh to another, we need to perform cross mesh resharding. One cross mesh resharding problem could be decomposed many "N-to-M data send task". Basically, one tensor could be decomposed into many tensor slices, each tensor slice originally has copy in some devices in mesh A, we want to send the tensor slice to some other devices in mesh B. Mesh A and Mesh B have no common devices. By "N-to-M data send task", I mean one data resides in N source devices and sends to M target devices. 

Current implementation is based on NCCL send, recv and allgather operators. Basically, using send/recv to make each target device get a part of the data, then do local allgather to make each target device get the whole data. Sender could be chosen by balancing load. This solution perform optimally under most cases, but does not solve all. 

**NCCL broadcast based cross mesh resharding**

Consider if we only use one device as the data sender, i.e. N=1, then the "1-to-M data send task" is exactly NCCL broadcast is doing. I check the NCCL broadcast implementation, it is also ring based pipeline algorithm as shown in the figure below, similar to allreduce implementation. That implementation allows total execution time to be much shorter. You could refer [NCCL-operator-rationale.pdf](https://github.com/Avkash/mldl/blob/master/pages/docs/books/NCCL-Woolley.pdf). 

<img width="1237" alt="image" src="https://user-images.githubusercontent.com/45677459/166100314-fd1588b7-14ce-45b7-be90-8fe445011f22.png">

When N is not 1, choosing arbitrary source device by load balancing to force N to be 1, could already achieve minimal cross host communication which is the bottleneck of speed. We could also create more broadcast tasks starting from different source devices and to different sets of adjacent target devices, so as to take advantage of concurrent execution. 

Theoretically, broadcast based implementation is optimal under all cases in terms of cross-host communication overhead. And ring-based implementation of NCCL broadcast with good load balancing could achieve even better/even optimal overall execution time. 

**Experiment results**

For single unit test by `tests/test_cross_mesh_resharding.py`, we could design resharding task on which broadcast based implementation is more than 4 times faster than send/recv+local allgather implementation as in the following task. Actually, we can design task on which broadcast based resharding is even much faster than send/recv+local allgather implementation, because we know current broadcast based resharding's shortcoming. 

[...]

PR #486, opened 2022-06-05
[FEATURE] Optimize order of shard tiles communication in cross mesh resharding

TarzanZhao:
### Background

Mentioned in [issue 416](https://github.com/alpa-projects/alpa/issues/416 
), one cross mesh resharding task is composed of many "N-to-M data send task" and each "N-to-M data send task" could be handled by NCCL send/recv or NCCL broadcast. In current implementation, one ray actor executes these send/recv/broadcast NCCL instructions one by one without concurrency for one cross mesh resharding task. That means we have to wait for the first NCCL instruction to finish before launching the second NCCL instruction. When the first instruction is blocked, all the following instructions will be blocked. Therefore the order of these NCCL instructions to be executed should be carefully decided. Here is an example showing how order will affect speed. 
image

I test this resharding example and my optimization of order indeed could speed up the overall performance. 
Here is the test code: src_mesh involves 2 hosts and dst_mesh involves 3 hosts. 
image
The experiment log clearly shows effectiveness of optimizing order.
image

On top of order, we should also decide which device should be used as sender for every task. Sender selection is about load balancing among different devices/hosts and order is about load balancing along time axis.

### My implementation

I define the above problem in *ReshardingLoadBalancingTaskSolver* and abstract it into a mathematically decent form 
by *AbstractedLoadBalancingTaskSolver*. I implement a search algorithm *AbstractedLoadBalancingTaskSolverSearchAlgo* for small-scale cases and a greedy algorithm *AbstractedLoadBalancingTaskSolverGreedyAlgo* for large-scale cases. The DFS based search algorithm is slow and could be improved by designing better pruning strategies. The greedy algorithm is not optimal and could be further improved. 

### Current Limitations to apply the features
1. Because the bottleneck of cross mesh resharding is usually at NIC which is host level. Thus we need at least 4 hosts in one resharding task(2 in src_mesh and 2 in dst_mesh) to possibly have concurrency. Current Alpa example only involves one host as src_mesh and one host as dst_mesh. The above example involves 5 hosts.  Such resharding cases never occur in current end-to-end training example of Alpa. That means this Feature could not give us immediately improvements now. Maybe if we have multiple NICs in one hosts, the features could play a more important role. 
2. Since the problem has been abstracted out and it is NP-hard, algorithmic solutions to it could be continuously designed and improved. Current DFS is slow (when we have 20 more hosts in one resharding) and Greedy algorithm could be hacked. 

We could add this feather in main branch when we need it later. I create this PR to keep Alpa collaborators updated and gather suggestions and requirements.


zhisbug:
Thank you. I'll try to finish reviewing this one by Thursday

zhuohan123:
@TarzanZhao @zhisbug Do we have any updates on this PR?

merrymercy:
closed due to inactivity

Issue #466, opened 2022-05-24
Model Checkpoint Improvements


merrymercy:
The current checkpoint design needs some improvements.

## Format
- The on-disk file names should be readable and deterministic instead of random numbers.
  For example, if we have two calls
  ```python
  save_checkpoint(path, target)
  save_checkpoint(path, target)
  ```
  The second call should fully overwrite the first call. However, in the current implementation with random numbers, we store two copies and the former will never be deleted or used.
- We should not store any absolute path

## Speed
Due to frequent [synchronization](https://github.com/alpa-projects/alpa/blob/a0d1d17ddc444924c27ef11816bc0963750570b9/alpa/device_mesh.py#L1417) and [compression](https://github.com/alpa-projects/alpa/blob/a0d1d17ddc444924c27ef11816bc0963750570b9/alpa/device_mesh.py#L287), the current save/load speed is very slow. We need to benchmark the performance against the peak bandwidth.

The peak bandwidth of load can be measured by running a jax program that reads np arrays from disk to local GPU memory. Ditto for save. When we save/load distributed arrays with multiple workers, we should also try to get a perfect linear scaling.
The possible ways to improve bandwidth include batching small arrays, batching ray calls, multithreading, and overlapping driver process with worker processes.

The micro-benchmark script can be put under `https://github.com/alpa-projects/alpa/tree/main/playground/alpa_micro_benchmark`

## Features
As a bonus, we want to support partial writing and partial reading. For example, we can save the checkpoints in two separate calls.
```python
# Write layer 1
param_tree_1 = {
  "1": {"kernel": array, "bias": array},
  "2": {"kernel": None, "bias": None},
}
save_checkpoint(path, param_tree_1)

# Write layer 2
param_tree_2 = {
  "1": {"kernel": None, "bias": None},
  "2": {"kernel": array, "bias": array},
}
save_checkpoint(path, param_tree_2)
```
Ditto for load.

## Minors
The aval for replicated arrays is the same 
https://github.com/alpa-projects/alpa/blob/a0d1d17ddc444924c27ef11816bc0963750570b9/alpa/serialization.py#L177

PKUFlyingPig:
I did some benchmark experiments for save/load. I chose three different methods for save/load: (1) flax's checkpoint functions (2) alpa's current checkpoint functions (no compression) (3) np.save/load. All the experiments are conducted both on EFS filesystem and local filesystem (results indicated that EFS can be the bottleneck).

[...]

PR #487, opened 2022-06-05
[RUNTIME] Model Checkpoint Improvements

PKUFlyingPig:
This PR improves alpa's model checkpoint functions according to #466.
# Format
- The on-disk file names are readable and deterministic.
I dfs the pytree and construct the filename according to the path from the root to leaf. 
For example: 
```
state = {"step": 1, "params": {"x": arr1, "b": arr2}, "list": [elem1, elem2]}
```
The paths for each element will be:
```
1 => state.step
arr1 => state.params.x
arr2 => state.params.b
elem1 => state.list.0
elem2 => state.list.1
```
- The paths stored in metadata are relative paths.

# Speed
All the benchmark results are in `alpa/playground/alpa_micro_benchmark/benchmark_dist_save_load.py`.
The peak bandwidth of load is measured by running a jax program that reads a 1GB numpy array from disk to local GPU memory. Ditto for save. The results are as followed:
```
np.save:                 save average run time: 0.8104 seconds, save average throughput:  9.8718 Gbps
np.load:                load average run time: 0.7327 seconds, load average throughput: 10.9179 Gbps
```
To achieve the peak bandwidth, I applied several optimizations: (1) I replaced tensorstore with np.save/load (2) batch the ray remote calls (3) write all the data into local disk instead of EFS.
To compare with the peak bandwidth, I shard a 1GB numpy array on a 1x4 mesh to get a DistributedArray, and save it to disk. Ditto for load. The results are as followed:
```
alpa.save:      save average run time: 0.8113 seconds, save average throughput: 9.8601 Gbps
alpa.load:     load average run time: 1.5235 seconds, load average throughput: 5.2512 Gbps
```
The alpa.load is two times slower than the peek bandwidth because the array is saved into several shards during alpa.save. To load the array, I need to read all the shards, construct the complete array and get the new shard according to the new shardingspec. I profile the code, and find [this assignment](https://github.com/PKUFlyingPig/alpa/blob/81855da44b3888e5506a11a365d479875be4a88f/alpa/serialization.py#L69) is which the overhead comes from (not np.load, but just the assignment). 
I also benchmarked the alpa.save/load on a 3GB MLP model which parallelled on a 2x4 mesh. The results are as followed:
```
alpa.save:  save average run time: 1.2081 seconds, save average throughput: 19.8683 Gbp
alpa.load: load average run time: 1.6163 seconds, load average throughput: 14.8510 Gbps
```
I may do more experiments to further improve the performance, but current version is much faster than the previous version. It should save/load the 170GB OPT model in 2-3 minutes.

### DIFF
--- a/alpa/device_mesh.py
+++ b/alpa/device_mesh.py
@@ -1401,7 +1401,7 @@ class DistributedArray:
         self._npy_value = None
 
     ##### distributed save/load #####
-    def save(self, ckpt_dir: str, local_cache_dir: Union[str, None]=None,  synchronized: bool=True):
+    def save(self, ckpt_dir: str, local_cache_dir: Union[str, None]=None):
         """
             Save one replica of the array to `ckpt_dir` distributedly.
 
@@ -1410,8 +1410,7 @@ class DistributedArray:
                 local_cache_dir: If not None, `ckpt_dir` should be a shared filesystem path, and this function 
                                  will return as soon as the shards have been saved to this local directory. 
                                  DaemonMoveWorkers will move these shards into `ckpt_dir` in the background.
-                synchronized: If set to False, it will return the ray futures. Or it will 
-                              wait untill all the shards have been saved to the disk.
+
         """
         one_replica_buffers = [
             self.remote_buffers[i] for i in self.one_replica_buffer_indices
@@ -1428,23 +1427,16 @@ class DistributedArray:
             else:
                 buf_refs_per_host[buf_ref.host_id].append(buf_ref.uuid)
                 indices_per_host[buf_ref.host_id].append(indice)
-        obj_refs = []
         for host_id, uuids in buf_refs_per_host.items():
             if len(uuids) > 0:
-                obj_refs.append(
-                    self.device_mesh.workers[host_id].save_buffers.remote(
-                        ckpt_dir, local_cache_dir, uuids, indices_per_host[host_id], self.shape))
-        if synchronized:
-            return None
-        else:
-            return obj_refs
+                self.device_mesh.workers[host_id].save_buffers.remote(
+                ckpt_dir, local_cache_dir, uuids, indices_per_host[host_id], self.shape)
 
     @classmethod
     def load(cls, path: str, aval: ShapedArray, device_mesh: PhysicalDeviceMesh,
-             sharding_spec: ShardingSpec, synchronized=True):
+             sharding_spec: ShardingSpec):
         """
             Load the data from `path` distributedly with `aval` and return a new DistributedArray
-            If set synchronized to False, it will return the futures.
         """
         # pylint: disable=import-outside-toplevel
         from alpa.mesh_executable import create_remote_buffer_refs
@@ -1463,19 +1455,12 @@ class DistributedArray:
                 buf_refs_per_host[buf_ref.host_id].append(buf_ref.uuid)
                 indices_per_host[buf_ref.host_id].append(indice)
                 device_ids_per_host[buf_ref.host_id].append(buf_ref.device_id)
-        obj_refs = []
         for host_id, uuids in buf_refs_per_host.items():
             if len(uuids) > 0:
-                obj_refs.append(
                     device_mesh.workers[host_id].load_buffers.remote(
                         path, uuids, indices_per_host[host_id],
-                        device_ids_per_host[host_id]))
-        loaded_array = DistributedArray(device_mesh, aval, sharding_spec, buf_refs, indices)
-        if synchronized:
-            ray.get(obj_refs)
-            return loaded_array
-        else:
-            return obj_refs, loaded_array
+                        device_ids_per_host[host_id])
+        return DistributedArray(device_mesh, aval, sharding_spec, buf_refs, indices)
 
     @property
     def one_replica_buffer_indices(self):
--- a/alpa/device_mesh.py
+++ b/alpa/device_mesh.py
@@ -279,10 +279,10 @@ class MeshHostWorker:
                      shard_indices: Sequence[Index],
                      device_ids: Sequence[int]):
         assert len(uuids) > 0
-        metadatas = list(filter(lambda fname: fname[:9] == ".metadata", os.listdir(ckpt_dir)))
+        metadatas = list(filter(lambda fname: fname.startswith(".metadata"), os.listdir(ckpt_dir)))
         # pylint: disable=import-outside-toplevel
-        from alpa.serialization import load_entire_arr
-        entire_arr = load_entire_arr(ckpt_dir, metadatas)
+        from alpa.serialization import load_sharded_array
+        entire_arr = load_sharded_array(ckpt_dir, metadatas)
         for index, uuid, device_id in zip(shard_indices, uuids, device_ids):
             data = entire_arr[index]
             self.put_buffer(uuid, device_id, data)
@@ -1401,7 +1401,7 @@ class DistributedArray:
                     self.device_mesh.workers[host_id].save_buffers.remote(
                         path, uuids, indices_per_host[host_id], self.shape))
         if synchronized:
-            return ray.get(obj_refs)
+            return None
         else:
             return obj_refs
 
@@ -1436,12 +1436,12 @@ class DistributedArray:
                     device_mesh.workers[host_id].load_buffers.remote(
                         path, uuids, indices_per_host[host_id],
                         device_ids_per_host[host_id]))
+        loaded_array = DistributedArray(device_mesh, aval, sharding_spec, buf_refs, indices)
         if synchronized:
             ray.get(obj_refs)
-            return DistributedArray(device_mesh, aval, sharding_spec, buf_refs, indices)
+            return loaded_array
         else:
-            return obj_refs, DistributedArray(device_mesh, aval, sharding_spec, buf_refs, indices)
-
+            return obj_refs, loaded_array
 
     @property
     def one_replica_buffer_indices(self):
--- a/alpa/device_mesh.py
+++ b/alpa/device_mesh.py
@@ -1454,8 +1454,11 @@ class DistributedArray:
         self._npy_value = None
 
     ##### distributed save/load #####
-    def save(self, path: str):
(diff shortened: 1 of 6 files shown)


zhuohan123 (changes requested):

zhuohan123:
In addition, how is resharding currently being handled?

alpa/device_mesh.py
```diff
@@ -1404,14 +1378,20 @@ def save(self, path: str):
         for host_id, uuids in buf_refs_per_host.items():
             if len(uuids) > 0:
                 obj_refs.append(
-                    self.device_mesh.workers[host_id].save_buffers_to_ts.remote(
+                    self.device_mesh.workers[host_id].save_buffers.remote(
                         path, uuids, indices_per_host[host_id], self.shape))
-        return ray.get(obj_refs)
+        if synchronized:
+            return ray.get(obj_refs)
```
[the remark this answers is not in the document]
PKUFlyingPig: Linter requires that I must return something since it will return the ray futures if synchronized is set to False. Maybe I should return None?

alpa/serialization.py
```diff
@@ -28,169 +26,134 @@
...
+    elif isinstance(tree, (tuple, list)):
+        for i, v in enumerate(tree):
+            _dfs_pytree(v, prefix + "." + str(i), paths)
+    elif tree is None:
+        return
     else:
-        spec['kvstore'] = {'driver': 'file', 'path': ckpt_path}
-    return spec
+        # Leaf node
+        paths.append(prefix)
```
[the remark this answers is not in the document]
PKUFlyingPig: Cite from [Jax's Doc](https://jax.readthedocs.io/en/latest/pytrees.html):
By default, pytree containers can be lists, tuples, dicts, namedtuple, **None**, OrderedDict. Other types of values, including numeric and ndarray values, are treated as leaves.
Since there is not way to further flatten the None container, I simply skip it.

alpa/serialization.py
```diff
@@ -28,169 +26,134 @@
...
 
-def ts_store(ckpt_dir, data: Union[np.ndarray, jax.xla.DeviceArray]):
-    ts_spec = get_ts_spec(ckpt_dir)
-    dtype = data.dtype
-    if dtype == jnp.bfloat16:
-        # Tensorstore uses 'bfloat16', not '<V2'.
-        dtype = 'bfloat16'
-    else:
-        dtype = np.dtype(dtype).str
+def _save_arr(ckpt_dir, arr):
```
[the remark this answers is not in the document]
PKUFlyingPig: I think `_save_unsharded_array` may be a better choice? This function is used to save a normal jax model (not parallelled by alpa yet) into the format which alpa can load.

alpa/device_mesh.py
```diff
@@ -1404,14 +1378,20 @@ def save(self, path: str):
...
             if len(uuids) > 0:
                 obj_refs.append(
-                    self.device_mesh.workers[host_id].save_buffers_to_ts.remote(
+                    self.device_mesh.workers[host_id].save_buffers.remote(
                         path, uuids, indices_per_host[host_id], self.shape))
-        return ray.get(obj_refs)
+        if synchronized:
+            return None
+        else:
+            return obj_refs
```
merrymercy, line 1386: You do not need to return futures here. If users want to sync or wait for this buffer. They can call `executable.sync()`, `device_mesh.sync_workers()` or `array.block_until_ready()`

alpa/device_mesh.py
```diff
@@ -1429,13 +1409,19 @@ def load(cls, path: str, aval: ShapedArray, device_mesh: PhysicalDeviceMesh,
...
+                obj_refs.append(
+                    device_mesh.workers[host_id].load_buffers.remote(
+                        path, uuids, indices_per_host[host_id],
+                        device_ids_per_host[host_id]))
+        loaded_array = DistributedArray(device_mesh, aval, sharding_spec, buf_refs, indices)
+        if synchronized:
+            ray.get(obj_refs)
+            return loaded_array
+        else:
+            return obj_refs, loaded_array
```
merrymercy, line 1424: No need to return anything

alpa/serialization.py
```diff
@@ -28,169 +25,137 @@
...
+                _save_unsharded_array(os.path.join(save_dir, arr_dir), x)
+            flat_metadata.append(arr_dir)
+        else:
+            flat_metadata.append(x)
+
+    metapath = os.path.join(nfs_dir, f"checkpoint_{step}")
+    metadata = tree_unflatten(target_tree, flat_metadata)
+    with open(metapath, "wb") as metafile:
+        metafile.write(msgpack.packb(metadata))
+    ray.get(obj_refs)
```
merrymercy, line 124: You do not need to call `ray.get` here. All tests should still be correct. Users can explicitly call sync outside this function by `executable.sync()`, `device_mesh.sync_workers()` or `array.block_until_ready()`

alpa/serialization.py
```diff
@@ -202,30 +167,36 @@ def restore_checkpoint(ckpt_dir: Union[str, os.PathLike], step: int, load_info:
...
+                arrays.append(dist_arr)
             flat_load_state.append(ReplicatedDistributedArray(meshes, arrays)) 
         else:
-            aval, mesh, spec = info.get_info()
-            flat_load_state.append(DistributedArray.load(path, aval, mesh, spec))
+            mesh, spec = info.get_info()
+            obj_ref, dist_arr = DistributedArray.load(os.path.join(ckpt_dir, path), info.aval, mesh, spec, False)
+            obj_refs.extend(obj_ref)
+            flat_load_state.append(dist_arr)
+    ray.get(obj_refs)
```
merrymercy, line 201: The same as save.

merrymercy (changes requested):
The new version looks much better. I only have some comments on ray synchronization.

alpa/device_mesh.py
```diff
@@ -83,13 +84,20 @@
...
-
+    def move(self, from_dir: str, to_dir: str):
+        os.makedirs(to_dir, exist_ok=True)
+        process_pool = []
+        for file in os.listdir(from_dir):
+            from_path = os.path.join(from_dir, file)
+            to_path = os.path.join(to_dir, file)
+            process_pool.append(subprocess.Popen(["mv", from_path, to_path], 
+                                            stdin=None, stdout=None, stderr=None))
+        for p in process_pool:
```
merrymercy: Does using multiprocessing here always make moving faster?

alpa/device_mesh.py
```diff
@@ -82,13 +82,27 @@
                                      ["broadcast_specs", "group_name"])
 
 
+class DaemonMoveWorker:
+    """A ray actor that moves local checkpoint into EFS in the background."""
```
merrymercy, line 86:
```suggestion
    """A ray actor that moves local checkpoint to the network file system in the background."""
```

Not everyone uses EFS

merrymercy (approved):

alpa/device_mesh.py
PKUFlyingPig: No, the bottleneck is the EFS bandwidth. I will change it to `shutil.move`.

alpa/serialization.py
```diff
@@ -28,191 +24,154 @@
...
+            elif isinstance(x, (np.ndarray, jax.xla.DeviceArray)):
+                _save_unsharded_array(arr_path, x)
+            flat_metadata.append(arr_dir)
+        else:
+            flat_metadata.append(x)
+
+    metapath = os.path.join(ckpt_dir, f"checkpoint_{step}")
+    metadata = tree_unflatten(target_tree, flat_metadata)
+    with open(metapath, "wb") as metafile:
+        metafile.write(msgpack.packb(metadata))
```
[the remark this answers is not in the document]
PKUFlyingPig: I tried `pickle` first, but it failed to serialize something maybe called `init_func` (I forgot the exact name) in trainState. Since flax uses msgpack in its `save_checkpoint` function, so I use it too.

alpa/device_mesh.py
```diff
@@ -715,11 +716,18 @@ def sync(self):
...
+    def shutdown(self, forced=False):
+        if not self.launched:
+            return
         self.sync()
         self.buffers.clear()
         self.executables.clear()
         self.distributed_client.shutdown()
+        # shutdown DaemonMoveWorker
+        if not forced:
+            ray.get(self.move_worker.shutdown.remote())
```
zhuohan123, line 728: Some comments here:
1. Rename `move_worker.shutdown` -> `move_worker.sync`. Because it's not shutting down the move_worker.
2. Instead of only syncing the save worker when shutdown, add an explicit `sync_move_worker` call to synchronize the sync worker and call `sync_move_worker` here.
3. Eventually, we can expose an API in `PhysicalDeviceMesh` that lets a user explicitly synchronize all the model saving workers. This might be useful in the missing unit test below.

tests/test_dist_save_load.py
```diff
@@ -93,26 +93,25 @@ def test_distributed_array_save_load(self):
...
+
+            # Load previously saved DistributedArray with a different shardingSpec
+            # [[0,1],          [[0,1],  [[0,1],
+            #  [2,3],  shard    [2,3]]   [2,3]]
+            #  [4,5],  ====>   [[4,5],  [[4,5],
+            #  [6,7]]           [6,7]]   [6,7]]
+            load_sharding_spec = logical_mesh.make_tile_spec(
+                global_input_data1, [0, 1], [0])
+            dist_load_data1 = DistributedArray.load(
+                tmpdir, jax.ShapedArray(global_input_data1.shape, jnp.int32),
```
zhuohan123, line 108: Can you add a test on local cache dir?

zhuohan123 (commented):
One more thing. We just updated our formatter and linter. Please run your code against the latest formatter and linter. Thanks!

PKUFlyingPig:
> One more thing. We just updated our formatter and linter. Please run your code against the latest formatter and linter. Thanks!

The main branch seems not to pass the linter?

zhuohan123 (approved):
LGTM

PR #489, opened 2022-06-06
Distributed weight initialization

merrymercy:
Now we can use `alpa.parallelize` to parallelize a weight initialization function.

## API

```python
def train_step(state, batch):

    def loss_func(params):
        out = state.apply_fn(params, batch["x"])
        return jnp.mean((out - batch["y"])**2)

    grads = alpa.grad(loss_func)(state.params)
    new_state = state.apply_gradients(grads=grads)
    return new_state

def create_state():
    model = Model()
    rngkey = jax.random.PRNGKey(0)
    params = model.init(rngkey, jnp.ones((1, input_dim)))
    tx = optax.adam(learning_rate=1e-2)
    return TrainState.create(apply_fn=model.apply, params=params, tx=tx)

batch = {
    "x": jnp.ones((batch_size, input_dim)),
    "y": jnp.ones((batch_size, output_dim)),
}

train_step = alpa.parallelize(train_step, method=ShardParallel(num_micro_batches=num_micro_batches))
create_state = alpa.parallelize(create_state, method=CreateStateParallel(train_step, batch))

state = create_state()
state = train_step(state, batch)
```

## Changes
- Add a `alpa.set_seed` for setting random number seed
- Implement the above APIs for ShardParallel and PipeshardParallel
- Clean up the compilation of PipeshardParallel
- Other small fixes

Depends on https://github.com/alpa-projects/tensorflow-alpa/pull/114


alpa/device_mesh.py
```diff
@@ -274,6 +276,10 @@ def get_exec_total_allocation_size(self, uuid: int):
     def get_exec_grad_sync_channel_ids(self, uuid: int):
         return self.executables[uuid].grad_sync_channel_ids
 
+    def set_runtime_random_seed(self, seed: int):
+        for d in self.local_devices:
+            d.set_seed(seed)
```
zhuohan123, line 281: If we set the same seed for all devices, will we generate all the same random numbers (e.g., same weights & dropout masks) on the same device? Same question for seeds on different hosts and different device meshes below.

alpa/create_state_parallel.py
```diff
@@ -0,0 +1,48 @@
...
+def compile_create_state_executable(fun, in_tree, out_tree_thunk, static_argnums,
+                                    donated_invars, batch_invars, train_step,
+                                    other_args, *avals):
+    # Trace to get jaxpr
+    jaxpr, out_avals, consts = pe.trace_to_jaxpr_final(fun, avals)
+    out_tree = out_tree_thunk()
+    state_aval = tree_unflatten(out_tree, out_avals)
+
+    executable = train_step.get_executable(state_aval, other_args)
+    placement_specs = executable.get_placement_specs()[0]
```
zhuohan123, line 23: Here do we assume that the output of `fun` should have exactly the same shape as `train_step`'s 0-th input argument? If so, we can add a document that explicitly says that.

zhuohan123 (approved):
LGTM in general. Left some comments.

alpa/pipeline_parallel/primitive_def.py
```diff
@@ -24,7 +24,7 @@
 def mark_pipeline_boundary():
     """Mark the boundary of pipeline layers. We reuse pipeline_marker for this
     functionality."""
-    return pipeline_p.bind(name="boundary", mark_type="boundary")
+    return pipeline_p.bind(0, name="boundary", mark_type="boundary")
```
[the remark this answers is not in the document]
merrymercy: If this primitive takes no input and returns no output, it will trigger a bug in jaxpr tracing. I just add a random number here to workaround the bug.

alpa/pipeline_parallel/primitive_def.py
```diff
@@ -112,8 +112,9 @@ def _pipeline_abstract_eval(*args, **kwargs):
 def _pipeline_xla_translation(c, *args, **kwargs):
     # TODO(yonghao): separate identity and marker in JAX
     if kwargs["mark_type"] == "hook":
-        return xla_identity(c, "hook", *args)
-    return xla_pipeline_marker(c, kwargs["mark_type"], kwargs["name"], *args)
+        ret = xla_identity(c, "hook", *args)
```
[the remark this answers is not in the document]
merrymercy: Sorry, this is my debug code. I will delete them

alpa/device_mesh.py
merrymercy: The current design:
1. We set the same "initial" random seed for all devices in all meshes.
2. When a device generates random states. It uses `actual_seed = initial ^ (device_id_in_a_mesh + mesh_id * 1024)`. Currently I only implemented `initial ^ device_id_in_a_mesh` part (see https://github.com/alpa-projects/tensorflow-alpa/blob/6741748e6f214dc8000a4bd24967df642852bb1a/tensorflow/compiler/xla/service/gpu/rng_thunk.cc#L27). I will add the `mesh_id` support when I implement this for `PipeshardParallel`

alpa/create_state_parallel.py
merrymercy: Will do

alpa/pipeline_parallel/primitive_def.py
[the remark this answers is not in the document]
zhuohan123: What exactly is the bug? We have been using this boundary for a long while but we didn't to meet any bug before.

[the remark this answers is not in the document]
merrymercy: I fixed it.

merrymercy:
@zhuohan123 @ZYHowell Please take another look. I finished the support for pipeshard parallel. The most critical part is these lines 
https://github.com/alpa-projects/alpa/blob/1d61371c1938bf4ec99f840d0a9978f3e7c90678/alpa/create_state_parallel.py#L80-L137
where I created an executable following given sharding specs and mesh assignments. For `eval_step`, we will do similar things.

alpa/create_state_parallel.py
```diff
@@ -0,0 +1,177 @@
...
+                 pipeshard_config: PipeshardConfig,
+                 placement_specs: Sequence[PlacementSpec],
+                 out_tree: Optional[PyTreeDef] = None,
+                 static_argnums: Optional[Sequence[int]] = None):
+        super().__init__(mesh_group=mesh_group,
+                         pipeshard_config=pipeshard_config,
+                         num_batch=1,
+                         out_tree=out_tree,
+                         static_argnums=static_argnums)
+        self.placement_specs = placement_specs
```
ZYHowell, line 39: I suggest we use different names for input and output placement specs. Like `get_placement_specs`->`get_input_placement_specs`. Otherwise this class has a `placement_specs` member and a `get_placement_specs` method(inherited from PipeshardParallel) but the two have different meaning

alpa/create_state_parallel.py
```diff
@@ -0,0 +1,176 @@
...
+
+        output_shardings = []
+        for var, spec in zip(outvars, placement_specs):
+            if isinstance(var, Var):
+                var2mesh[var] = spec.mesh_ids[0]
+            output_shardings.append(spec.sharding_specs[0])
+
+        num_meshes = len(executable.mesh_group)
+
+        propagate_mesh_assignment(jaxpr, var2mesh, eqn2mesh)
```
ZYHowell, line 120: Apply Gradient in Pipeshard parallel's compile executable has a likewise function. I'll try to merge the two and move to utils.py

ZYHowell (approved):
LGTM. Only few comments.

alpa/shard_parallel/compile_executable.py
```diff
@@ -138,8 +139,9 @@ def shard_parallel_internal_gradient_accumulation(
     """Compile a gradient accumulation executable with auto-sharding pass."""
     # pylint: disable=unused-argument
     # Split the batch dimension
-    closed_jaxpr, avals, _ = trace_jaxpr_with_micro_batch(
+    closed_jaxpr, _ = trace_jaxpr_with_micro_batch(
         fun, batch_invars, num_micro_batches, raw_avals)
+    avals = closed_jaxpr.jaxpr.invars
```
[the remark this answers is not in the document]
merrymercy: Actually, these variables are never used in `run_auto_sharding_pass`. Let me delete these unused arguments in `run_auto_sharding_pass`

alpa/create_state_parallel.py
merrymercy: Sounds good. I renamed `get_placement_specs` to `get_input_placement_specs` and renamed `placement_specs` in `CreateStateExecutable` to `target_placement_specs`

alpa/create_state_parallel.py
merrymercy: They have a slight difference:
- in apply_grad, we can replicate the equations to favor recomputation over communication. 
- in create_state, due to our use of stateful random number generation, replicated computation can generate different results on different device meshes. So we cannot replicate equations. We have to use send/recv to send the temp results among meshes. To avoid the deadlock, we also need to fix an order.

merrymercy:
Comments are addressed

PR #491, opened 2022-06-06
[FEATURE] Support beam sample

ZYHowell:
This PR adds reorder cache function to support the beam sample. It modifies some apis in `opt_serving/model/wrapper.py` to add config for beam number.

### DIFF
--- a/alpa/device_mesh.py
+++ b/alpa/device_mesh.py
@@ -47,12 +47,11 @@ import ray
 from alpa import mesh_profiling
 import alpa.collective as col
 from alpa.global_env import global_config
-from alpa.parallel_plan import StagePlan
 from alpa.monkey_patch import set_override_backend
 from alpa.shard_parallel.auto_sharding import (AutoShardingOption,
                                                LogicalDeviceMesh,
                                                run_spmd_partitioner_pass)
-from alpa.parallel_plan import PlacementSpec
+from alpa.parallel_plan import PlacementSpec, StagePlan
 from alpa.timer import timers
 from alpa.util import (benchmark_func, list_gpu_info, OrderedSet,
                        update_jax_platform, is_ray_node_resource,
--- a/alpa/device_mesh.py
+++ b/alpa/device_mesh.py
@@ -1552,7 +1552,7 @@ class DistributedArray:
                                                    self.device_mesh.num_devices)
 
             as_option = AutoShardingOption()
-            strategy_config = StagePlan(global_config.build_random_seed,
+            strategy_config = StagePlan(global_config.compile_random_seed,
                                         self.device_mesh.shape, 1 << 60,
                                         as_option.all_reduce_threshold, None,
                                         -1)
--- a/alpa/device_mesh.py
+++ b/alpa/device_mesh.py
@@ -32,7 +32,7 @@ from typing import Any, List, Union, Sequence, Tuple, Optional
 import jax
 from jax import core, xla, device_put
 from jax._src.api import ShapeDtypeStruct
-from jax._src.lib import xla_bridge as xb, xla_extension as xe
+from jax._src.lib import xla_bridge as xb, xla_client as xc, xla_extension as xe
 from jax._src.tree_util import tree_leaves
 from jax.abstract_arrays import array_types
 from jax.core import ShapedArray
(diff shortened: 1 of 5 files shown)


examples/opt/model/test_text_gen.py
```diff
@@ -193,12 +206,13 @@ def inference_func(input_ids, past_key_values, output_attentions=False,
...
     parser.add_argument("--device", type=str, default="cuda")
     parser.add_argument("--dummy", action="store_true")
+    parser.add_argument("--num-beams", type=int, default=1)
     args = parser.parse_args()
 
     tokenizer = GPT2Tokenizer.from_pretrained("facebook/opt-125m")
 
     tic = time.time()
-    model = get_model(args.model, args.device, args.dummy)
+    model = get_model(args.model, args.device, args.dummy, args.num_beams)
```
zhisbug, line 215: this way of passing `num_beams` seems not aligned with our current interface

examples/opt/model/test_text_gen.py
```diff
@@ -139,22 +151,23 @@ def inference_func(input_ids, past_key_values,
...
+        inference_func_config.num_beams = num_beams
         print(inference_func_config)
 
     elif "alpa/opt" in model_name:
         alpa.init()
         num_pp_stages = max(2, alpa.get_global_cluster().num_hosts)
 
         name = model_name.split("-")[1].upper()
-        config = get_config(name, num_pp_stages=num_pp_stages)
+        config = get_config(name, num_pp_stages=num_pp_stages, batch_size=num_beams)
```
zhisbug, line 162: `batch_size` = num_beams?

examples/opt/model/test_text_gen.py
zhisbug: we must pass a beam here when compile?

ZYHowell: Beam size determines input tensor's shape and thus the pipeshard executable

examples/opt/model/test_text_gen.py
ZYHowell: If we need to support real batch size, the argument should be `batch_size=num_beams * batch_size`

zhuohan123: I think Yonghao's current way makes sense. `config.batch_size` is a feature of the model that defines the size of the batch dimension of the input, which would be `num_beams * batch_size` with beam search.

alpa/device_mesh.py
```diff
@@ -48,12 +48,16 @@
 from alpa import mesh_profiling
 import alpa.collective as col
 from alpa.global_env import global_config
+from alpa.parallel_plan import StagePlan
```
merrymercy, line 51: move to L53.

examples/opt_serving/model/opt_model.py
```diff
@@ -529,25 +529,28 @@ def init_model_aval(config):
     return model, params
 
 
-def init_cache_aval(config, batch_size):
+def init_cache_aval(config):
```
merrymercy, line 532: revert all changes of `init_cache_aval`. Let us keep `batch_size` as an argument because we may use multiple batch sizes for the same model config

examples/opt_serving/model/opt_model.py
```diff
@@ -529,25 +529,28 @@ def init_model_aval(config):
...
-                                 config.dtype),
+                                 dtype),
             jax.core.ShapedArray((batch_size,), jnp.int32),
         )
         all_cache.append(layer_cache)
     return tuple(all_cache)
 
 
-def init_cache_np(config, batch_size):
+def init_cache_np(config):
```
merrymercy, line 552: revert all changes

merrymercy (changes requested):
LGTM. Let us merge this soon. Since beam search introduces significant overhead, can we turn it off by default?

examples/opt_serving/model/opt_model.py
ZYHowell: Then shall we remove `batch_size` in `config`? Since we have a `batch_size` outside the config, it is weird to have another.

ZYHowell:
@zhisbug I test `textgen_demo.py` with `num_beams=2` on 125M and 2.7B models.
To support beam sample, the `batch_size` for `get_pipeshard_executable` will be greater than 1 so I move the check `auto_regressive && num_batch == 1` out of the function. All other changes do not affect previous cases that `num_beam=1` as default.

merrymercy:
@ZYHowell  Feel free to merge when it is ready

zhisbug:
@vinlnx see the above perf number

Issue #492, opened 2022-06-07
Get rid of Numba dependency


zhisbug:
If possible, can we want to get rid of the numba dependency? 
For now, it requires a special version of numpy, which makes the installation procedures twisted.

PR #493, opened 2022-06-07
[FIX] Make numba optional

zhuohan123:
Fixing #492


zhuohan123:
I tested multiple typical setups in our benchmark. And the result is that there is a ~250x speed difference for results with and without numba jit. The setting in our paper takes 1.65s, and the time will be ~400s without numba. I think numba is highly recommended but we can still run without numba.

If numba is really bothering, we can reimplement the DP in cython, which should be much more popular than numba.

PR #511, opened 2022-06-14
adding pjit

### LINKED ISSUE #505: Adding the `pjit` in the comparison
Some people are more familiar with using model parallel via [`pjit`](https://github.com/google/jax/blob/main/jax/experimental/pjit.py). What about adding one more rows [here](https://github.com/alpa-projects/alpa/blob/main/docs/gallery/tutorials/alpa_vs_pmap.py#L46-L52)?

JiahaoYao:
fix  #505


zhuohan123 (approved):

PR #528, opened 2022-06-18
Add RNG seed setting to other PT tests

yf225:
As titled.


yf225:
Ah I saw we have another commit addressing this, please feel free to close this as needed :)

PR #530, opened 2022-06-20
Fix Zhen model implementation

merrymercy:


merrymercy:
cc @yf225

yf225 (commented):
Thanks so much!

PR #533, opened 2022-06-21
[BENCHMARK] Add benchmark scripts for inference

PKUFlyingPig:
Configuration:
- `alpa/benchmark/alpa/suite_gpt_infer.py`: grid search configuration (BS, #B, #PP, #DP, #OP ... ...)
- `alpa/benchmark/alpa/exp_infer_gpt.sh`: 
    - `result_name`: output file name
    - run_experiment #Host #Devices_per_host

Usage:
```
bash exp_infer_gpt.sh
```


zhuohan123:
> I'll review -- but what's the difference between this script and this one? https://github.com/alpa-projects/alpa/blob/opt/examples/opt_serving/benchmark/benchmark_text_gen.py
> 
> Should we merge them?

One is autoregressive and one is not. Let's keep them separate for now.

zhisbug:
@zhuohan123 in the latest script, there is a mode to switch to perform forward-like computation (only for benchmarking usage).

See https://github.com/alpa-projects/alpa/blob/opt/examples/opt_serving/benchmark/benchmark_text_gen.py#L13

zhuohan123:
OK I see, I believe they should be merged, but not necessarily in this PR. I think we can move `benchmark_text_gen.py` under `benchmark/alpa` and merge it with the benchmark codes in this PR. Right now the main differences between the two are as below:

`benchmark_text_gen.py`:
1. Benchmark support for both autoregressive and non-autoregressive decoding.
2. Focus on language models.

This PR: 
1. Only support non-autoregressive cases. 
2. Support for providing parallel strategy config. 
3. Can be used as a template for evaluating other models (e.g., resnet).

zhuohan123 (changes requested):
Didn't see anything particularly wrong. Comments are mainly related to style.

zhuohan123 (approved):
Please run the formater before merging.

PR #535, opened 2022-06-22
[PERF] offload rng computation in remat

ZYHowell:
In `jax.remat`, constant values and random numbers are generated in the forward part and stored until the backward part. An example is [this](https://gist.github.com/ZYHowell/96e31b8e43ec37a9ddfaac4aa1a559aa). To reduce memory consumption, we remat this part as well. This PR:
- Move `offload_remat` to `util.py` as an essential part after trace jaxpr in `process_remat`.
- Store the rng result and reuse it as an input in the offloaded part,
- Add test cases for remat with rng seed.


alpa/util.py
```diff
@@ -621,13 +625,197 @@ def clone_jaxpr(closed_jaxpr: ClosedJaxpr,
...
+
+get_seed_p = Primitive("alpa_get_seed")
+set_seed_p = Primitive("alpa_set_seed")
+_seed_shape = xe.Shape.array_shape(np.int32, (1,))
+_seed_aval = ShapedArray((1,), np.int32)
+
+
+def _get_seed_translation(c, *args, **kwargs):
+    return xc.ops.CustomCall(c,
+                             b"alpa$get-seed",
```
merrymercy, line 650:
```suggestion
                             b"alpa$get-state",
```

ZYHowell:
cc @merrymercy ready for review

PR #544, opened 2022-06-25
[Docs] Quote ".[dev]" in installation guide 

jiaodong:
I'm following build from source guide in the docs:https://alpa-projects.github.io/install.html#install-from-source on zsh and it will complain `zsh: no matches found: .[dev]` 

Adding this quote helps to make installation successful.


zhisbug:
@jiaodong  Thanks for the contribution!

PR #555, opened 2022-06-29
[Feature] OPT weight conversion scripts, more detailed procedures

zhisbug:
- [ ] scripts for converting weights
- [ ] using rst instead md
- [ ] upload md5


larekrow:
Hi @zhisbug, can I check whether the consolidated weights after running `step_2_consolidate_992_shards_to_singleton.py` would be the same (or similar enough) as the output after running Metaseq's [reshard the shards steps](https://github.com/facebookresearch/metaseq/blob/main/projects/OPT/download_opt175b.md#reshard-the-shards)?

Was hoping if that were the case, we could use the Metaseq's `reshard_mp.py` on OPT-IML-175B (given as 16 TPs) followed by Alpa's `step_3_convert_to_numpy_weights.py` to run OPT-IML-175B with Alpa.

Issue #504, opened 2022-06-11
[PERF][REFACTOR] Use one uuid for all buffers of a tensor


ZYHowell:
## Background and Motivation

Currently, Alpa assigns a UUID for each remote device buffer, and thus a [distributed tensor](https://github.com/alpa-projects/alpa/blob/ef6df24f1c6603ce40ed78cf0837c2bee5ba64e7/alpa/device_mesh.py#L1382) with multiple remote buffers has multiple UUIDs and `RemoteBufferRef`s. Creating and deleting these objects is costly. However, since all operations are on the tensor level and each mesh worker executes SPMD programs, only one UUID is enough for a tensor. Using one UUID for all buffers of a tensor can reduce the runtime overhead by a constant factor. This is also what [Pathways](https://arxiv.org/pdf/2203.12533.pdf) does (in sec 4.2 client, they use sharded tensor equivalent to Alpa's DistributedArray).

## TODO
  - [x] Change the type of `MeshHostWorker.buffers`: From `Dict[int, DeviceArray]` to `Dict[int, Sequence[DeviceArray]]`.
    - Change the implementation of buffer-related methods in `MeshHostWorker`.
    - Move codes in `RemoteBufferRef` to `DistributedArray`.
    - Change return value of functions in `shard_arg_handlers`.
  - [x] ShardParallel: Rewrite `DriverExecutable.launch_on_driver` and `WorkerExecutable.execute_on_worker`'s pre- and post-process.
  - [x] CrossMeshResharding: Refactor cross-mesh resharding tasks' API.
    - Replace `buf_uuids` by `device_indices` in `MeshHostWorker`'s related functions.
  - [x] PipeshardParallel: Refactor `PipelineInstruction`.

PR #556, opened 2022-06-29
[Refactor] Use one uuid for all buffers of a logical tensor

ZYHowell:
https://github.com/alpa-projects/alpa/issues/504


alpa/device_mesh.py
```diff
@@ -154,60 +159,98 @@ def __init__(self, server_address: str, num_hosts: int, host_id: int,
...
-        self.buffers[uuid] = (self.backend.buffer_from_pyval(
-            data, self.local_devices[device_id]))
-
-    def put_buffers(self,
-                    uuids: Sequence[int],
-                    device_ids: Sequence[int],
-                    datas: Sequence[np.ndarray],
-                    num_batch=1,
-                    batch_dim=0):
+    def put_tensor(self,
```
merrymercy, line 162: We have several names "buffer", "array", and "tensor". In the old codebase, we use "array" for a high-level logical array and "buffer" for a low-level per-device buffer. Jax codebase never uses "tensor".

In the new codebase, let me suggest a naming scheme:
1. Following the jax convention, never use "tensor"
2. Rename `RemoteBufferRef` -> `RemoteArrayRef`.
3. Do not change the name of `put_buffer`, `get_buffer` because they still work on low-level per-device buffers

merrymercy (changes requested):

alpa/mesh_executable.py
```diff
@@ -322,15 +247,11 @@ def launch_on_driver(self, *args, **kwargs):
         timers(self.shard_args_timer_name).stop()
 
         if isinstance(physical_mesh, DistributedPhysicalDeviceMesh):
-            # Shape: (num_hosts, num_args, num_devices_per_host)
-            input_uuids = (get_uuid_np_array(input_bufs).reshape(
-                (len(args), num_hosts,
-                 num_devices_per_host)).transpose([1, 0, 2]))
+            # Shape: (num_args,)
+            input_uuids = [ref.uuid for ref in input_bufs]
```
merrymercy, line 251:
```suggestion
            input_uuids = np.array([ref.uuid for ref in input_bufs])
```

alpa/mesh_executable.py
```diff
@@ -732,57 +630,38 @@ def launch_on_driver(self, *args):
         timers(self.shard_args_timer_name).stop()
 
         if isinstance(physical_mesh, DistributedPhysicalDeviceMesh):
-            # Shape: (num_hosts, num_args, num_devices_per_host)
-            first_batch_uuids = (get_uuid_np_array(first_batch_bufs).reshape(
-                (len(first_batch_bufs), num_hosts,
-                 num_devices_per_host)).transpose([1, 0, 2]))
+            # Shape: (num_args,)
+            first_batch_uuids = [ref.uuid for ref in first_batch_bufs]
```
merrymercy, line 634: similarly, wrap all arguments of `ray.remote` with `np.array`

alpa/device_mesh.py
```diff
@@ -154,60 +159,98 @@ def __init__(self, server_address: str, num_hosts: int, host_id: int,
...
                 split_datas.extend(split_buffers)
             datas = split_datas
-        for uuid, device_id, data in zip(uuids, device_ids, datas):
+        tensors = [([None] * self.num_devices) for _ in range(num_batch)]
+        for i, data in enumerate(datas):
             if data.dtype == np.int64:
                 data = data.astype(np.int32)
-            self.buffers[uuid] = (self.backend.buffer_from_pyval(
+            device_id = i // num_batch
+            batch_id = i % num_batch
```
merrymercy, line 182: use `divmod`

alpa/device_mesh.py
```diff
@@ -1099,29 +1201,23 @@ def shard_args_to_bufs(self, shard_indices: Sequence[Sequence[Index]],
                     slow_path = True
```
merrymercy, line 1201: update L1199:
```python
                    ret_bufs.append([arg.remote_ref])
```

alpa/device_mesh.py
```diff
@@ -1146,9 +1242,9 @@ def shard_args_to_arrays(self, avals: Sequence[ShapedArray],
                              args: Sequence[np.array]):
         arrays = []
         for i in range(len(avals)):
-            buffers = _shard_array(args[i], self, shard_indices[i])
+            remote_ref = _shard_array(args[i], self, shard_indices[i])[0]
```
merrymercy, line 1245: This extra `[0]` seems wired. Any way to eliminate it?

merrymercy (changes requested):

merrymercy:
Despite the naming issue. Other major changes look very good! I am glad to see how this simplifies a lot of runtime dispatching code.

alpa/device_mesh.py
```diff
@@ -1258,12 +1402,12 @@ def __init__(self,
                  device_mesh: PhysicalDeviceMesh,
```
merrymercy, line 1402: update the docstring of this class

alpa/device_mesh.py
ZYHowell: `put_buffer` cannot work on per-device buffers. `get_buffer` does so because there might be some redundancy due to the sharding spec, so we only get exactly one replica. currently `put_buffer` does not consider the exactly-one-replica stuff. We might do this optimization later so I reverted the name too.

alpa/device_mesh.py
ZYHowell: We used to have some `if` to return a single value instead of a list when the input is not iterable or some arg is 1. but this feature makes some code more complicated now.
For example, for the case that `num_arg == 1`, we need to add code like `ret = np.array([ret])`. Previously we do not have the problem because when `num_arg == 1`, if `num_devices != 1`(this is a usual case), the special branch will not be hit.
I didn't find a perfect balance between the `[0]` and `if num == 1: ret = [ret]`. Do you have some suggestions?

TarzanZhao (commented):
This is the first time I review code. I mainly read the code that I am familiar with in `worker_nccl_util_xla.py`, `worker_nccl_util_cupy.py`, `cross_mesh_resharding.py` `device_mesh.py`, they look good to me. BTW I see many methods/functionality being deleted. I am not sure whether they will no longer be used or not, like put_non_zero_buffer and put_buffer in `device_mesh.py`.

TarzanZhao:
> 

OK, get it!

PR #571, opened 2022-07-01
[FEATURE] Allow num_return_sequences > 1 (#491)

vinlnx:
An extension to 14b31536cecdccfea8e043d538f4b034be1063a8 to enable the return on multiple outputs.

### DIFF
--- a/examples/opt_serving/model/wrapper.py
+++ b/examples/opt_serving/model/wrapper.py
@@ -36,7 +36,7 @@ class InferenceFuncOutput(ModelOutput):
 class InferenceFuncConfig:
     """Implements a minimal config class for using huggingface's generator.
 
-    Note: these paramerers might be overwritten by model.generate(**kwargs).
+    Note: these parameters might be overwritten by model.generate(**kwargs).
     """
     bos_token_id: int = 0
     num_beams: int = 1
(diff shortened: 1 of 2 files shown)


merrymercy (approved):

merrymercy:
Thanks, @vinlnx. This is merged.

examples/opt_serving/model/wrapper.py
```diff
@@ -287,6 +289,15 @@ def get_model(model_name: str,
     if not dummy:
         assert os.path.exists(path), f"No such file or directory: '{path}'"
 
+    # figure out the actual input size
+    if do_sample:
+        expand_size = batch_size * num_beams * num_return_sequences
+    else:
+        if num_return_sequences > num_beams:
+            raise ValueError(
```
pnpnpn, line 297: Is this needed? For example, if I want to `generate num_return_sequences=4`, why does this need to take 16 for `expand_size`? 

I was able to generate different samples with `num_beams=1` and `num_return_sequences=4` even when I commented this out. 

The `num_return_sequences=3` was flaky though. I suppose it's related to power-of-2.

vinlnx: So when `num_beams > 1` and `do_sample == True`, the beam scorer does what's called beam-search multinomial sampling (https://github.com/huggingface/transformers/blob/main/src/transformers/generation_utils.py#L1381).

You can definitely set `num_beams=1` and just do sampling, but if you set `num_beams > 1`,  `num_return_sequences > 1`, and `do_sample == True`, it will just return the same output `num_return_sequences` times.

PR #576, opened 2022-07-02
[PERF] update index select for beam search

ZYHowell:
The reason why index select was slow is the reordered cache is batch vars, but it was not set as skip resharding.


examples/opt_serving/model/wrapper.py
```diff
@@ -130,35 +139,74 @@ def _reorder_cache(self, past, beam_idx):
...
+                dim = 0
+                avals = [ary.aval for ary in arys]
+                specs = [ary.sharding_spec for ary in arys]
+                executable = get_index_select_mesh_executable(
+                    avals, specs, mesh_idx, dim, device_mesh,
+                    [False] * len(avals))
+                self.index_select_executables[device_mesh] = executable
+            ret = executable(*arys, mesh_idx)
+            for v in ret:
+                v.skip_shard_args_check = True
```
merrymercy, line 178: use `wrapper.py::set_skip_shard_args_check`.

merrymercy (approved):

merrymercy:
What's the perf number? Is there any room for improvement?

merrymercy (changes requested):
Do not merge. I will do some refactor

PR #587, opened 2022-07-04
[FEATURE] Serialize Parallel Plan

merrymercy:
Now we can serialize a parallel plan and reuse it. Currently, it only supports pickle, which is not good for backward compatibility. We can design a better format later.

```python
plan = executable1.get_parallel_plan()

with open("tmp_plan.pkl", "wb") as fout:
    pickle.dump(plan, fout)
with open("tmp_plan.pkl", "rb") as fin:
    plan = pickle.load(fin)

p_train_step = parallelize(train_step, method=plan_to_method(plan))
executable2 = p_train_step.get_executable(state, batch)
```

### DIFF
--- a/alpa/monkey_patch.py
+++ b/alpa/monkey_patch.py
@@ -10,7 +10,7 @@ from jax._src.lib import xla_client as xc
 from jax._src.lib.mlir.dialects import mhlo
 from jax._src.lib.xla_bridge import get_backend as default_get_backend
 from jax.core import Primitive
-from jax.interpreters import partial_eval as pe
+from jax.interpreters import partial_eval as pe, pxla
 from jax.interpreters import xla, mlir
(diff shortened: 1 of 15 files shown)


alpa/monkey_patch.py
```diff
@@ -228,6 +229,64 @@ def _remat_translation_rule(ctx,
...
+
+    # pylint: disable=unnecessary-dunder-call
+    self.__init__(
+        sharding=sharding,
+        mesh_mapping=mesh_mapping,
+    )
+
+
+setattr(pxla.ShardingSpec, "__getstate__", sharding_spec_getstate)
+setattr(pxla.ShardingSpec, "__setstate__", sharding_spec_setstate)
```
merrymercy, line 287: @ZYHowell  Now we can pickle `ShardingSpec`

PR #590, opened 2022-07-05
[FIX] do allgather with GShard generated code in XLA

ZYHowell:
Alpa's allgather algorithm only supports gathering a continuous buffer due to NCCL API. Besides, the performance is bad when the to-be-gathered slice is not continuous because it runs the slow path in `recv_tile`. This design aims to do in-place AllGather when the to-be-gathered slice is continuous. The in-place AllGather reduces a copy compared with not in-place version. However, the gain from a copy is negligible, but the in-place AllGather is complicated and limited.

This PR moves it to a normal AllGather: the `recv_tile` stores the result continuously, then the mesh runs AllGather that is not in-place. We simply compile an XLA executable and let GShard insert AllGather.

Experiment shows that in-place AllGather does not have significant improvement: for testcase`model_spec=GPT-6.7B, batch_size=512, #mb=64, parallel_config=(1,4,4)`, the in-place AllGather does not have any copy, but the perf is: No AllGather/In-place AllGather/XLA AllGather(TFLOPS(ckpt)) = 46/59/59.


alpa/util.py
```diff
@@ -549,6 +549,32 @@ def compile_concatenate(backend, mesh_shape, sharding_spec, batch_size,
     return hlo_proto
 
 
+def compile_allgather(shape, dtype, src_spec, dst_spec, num_devices):
```
merrymercy, line 552: Document these functions

merrymercy (commented):
1. remove `use_scatter_gather` in `global_env.py` to `use_local_allgather`
2. We can skip the xla_nccl test `test_xla_nccl` for now because xla nccl backend is not used currently.

merrymercy:
@ZYHowell Let us try to get this merge soon

ZYHowell:
> @ZYHowell Let us try to get this merged soon

All comments are addressed but @zhisbug may want to review this PR so I'll merge after his review.

merrymercy (approved):

PR #603, opened 2022-07-08
[OPT] Unify forward-only and autoregressive benchmark

comaniac:
This PR unifies two benchmark modes for OPT. Specifically:
1. Forward-only mode now takes cache.
2. Forward-only mode will use `test_prompts` if the user-specified decoder length matches any of them.
3. Unify the wrapper interface `__call__` by moving the prompt decomposing logic to autoregressive specific inference function.

I'll further debug the cache issue with this PR.

cc @merrymercy

### DIFF
--- a/examples/opt_serving/benchmark/benchmark_text_gen.py
+++ b/examples/opt_serving/benchmark/benchmark_text_gen.py
@@ -59,8 +59,8 @@ if __name__ == "__main__":
     args = parser.parse_args()
 
     # Some global params
-    warmup_iters = 2
-    n_iters = 0
+    warmup_iters = 5
+    n_iters = 10
     global_config.pipeline_sync_for_timer = True
     global_config.shard_parallel_sync_for_timer = True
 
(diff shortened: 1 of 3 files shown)


merrymercy (approved):

merrymercy:
@zhisbug If you don't have any comments, I will merge this in 6 hours

merrymercy:
@zhisbug If you don't have any comments, I will merge this in 6 hours
@comaniac Please fix the conflicts.

It seems we should also add some unit test for OPT

merrymercy:
It seems this is not compatible with #623. I don't know how we should merge this

comaniac:
We could merge the batch inference maybe as this PR itself is mainly just for code cleaning.

zhisbug:
@comaniac I have merged the batched inference PR; could you rebase and see if everything works well? Then we can merge this one

comaniac:
We discussed this last week. Please go ahead and merge #623, and I'll implement another PR for multi-executable with 1-D OPT support later, as we are not bucketing based on sequence length anymore. This PR will eventually be closed.

merrymercy:
Close Stale Issues

PR #607, opened 2022-07-10
[Doc] Update path to jaxlib build

ludgerpaehler:
We are already in the alpa folder at this point, we hence need to move one step shorter. The instructions to runs installation verification tests are also not working as is, and the instructions have been amended as such.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -130,7 +130,7 @@ You can check the installation by running the following commands from the root o
 .. code:: bash
 
   ray start --head
-  python3 tests/test_install.py
+  python3 -m alpa.test_install
 
 [Optional] PyTorch Frontend
 -------------------------------------
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -109,7 +109,7 @@ Method 2: Install from Source
 
   .. code:: bash
 
-    cd alpa/build_jaxlib
+    cd build_jaxlib
     python3 build/build.py --enable_cuda --dev_install --tf_path=$(pwd)/../third_party/tensorflow-alpa
     cd dist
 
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -125,12 +125,12 @@ Method 2: Install from Source
...
 Check Installation
 ------------------
-You can check the installation by running the following commands.
+You can check the installation by running the following commands from the root of the repository.
 
 .. code:: bash
 
   ray start --head
-  python3 -m alpa.test_install
+  python3 tests/test_install.py
```
merrymercy, line 133: `python3 tests/test_install.py` is deprecated.
`python3 -m alpa.test_install` is recommended because it also works if people instal by wheel

docs/install.rst
```diff
@@ -109,7 +109,7 @@ Method 2: Install from Source
 
   .. code:: bash
 
-    cd alpa/build_jaxlib
+    cd build_jaxlib
```
merrymercy, line 112: This change looks good

merrymercy (changes requested):

docs/install.rst
```diff
@@ -125,12 +125,12 @@ Method 2: Install from Source
 
 Check Installation
 ------------------
-You can check the installation by running the following commands.
+You can check the installation by running the following commands from the root of the repository.
```
merrymercy, line 128:
```suggestion
You can check the installation by running the following commands.
```

docs/install.rst
merrymercy, line 133:
```suggestion
  python3 -m alpa.test_install
```

merrymercy (changes requested):

PR #608, opened 2022-07-11
[OPT] Fix attention bias when cache is used

comaniac:
The current OPT model generates attention bias with shape `[1, 1, 1, max_length]`. However, since its shape is `[batch_size, heads, input_length, max_length]`, the generated bias is incorrect when `input_length > 1`.

This PR fixes this issue. Note that the correct logic could be:

```python
attention_bias = jnp.expand_dims(
    jnp.triu(jnp.full(
    (num_updated_cache_vectors, max_length), -1e10), cache_index + 1), (0, 1))
```

However, we cannot use this implementation, because the 2nd argument of `triu` (i.e., k-th diag) has to be static. Alternatively, we extend the existing solution with 2 `arange`s.

Meanwhile, this PR also improves the `test_cache` to test both cases (whether or not decompose the input prompt) against the one without cache.

cc @zhuohan123

### DIFF
--- a/examples/opt_serving/model/opt_model.py
+++ b/examples/opt_serving/model/opt_model.py
@@ -157,8 +157,6 @@ class OPTSelfAttention(nn.Module):
                     jnp.full((query_states.shape[1], key_states.shape[1]),
                              -1e10), 1), (0, 1))
         else:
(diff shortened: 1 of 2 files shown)


zhuohan123 (approved):

PR #617, opened 2022-07-20
[INFERENCE] Get inference output from remoteBuffers to driver asynchronously in Pipeline Parallel.

PKUFlyingPig:
During inference, it is common for users to get the inference output immediately. Below is an example scenario:
```
for batch in batches:
    loss = infer_step(params, batch)  # return a DistributedArray
    print(loss._value)  # block until the last infer_step call finishes
```
However, in Alpa's current implementation, the `Distributed._value` property is a synchronous call, which will block until the last  `infer_step` function call finishes (completes its last pipeline stage). And only after that, the next `infer_step` can be called. So there is always only one batch in the pipeline, losing all the performance gain from pipeline parallel.

A straightforward workaround may be like this:
```
losses = [ ]
for batch in batches:
    loss = infer_step(params, batch)  # return a DistributedArray
    losses.append(loss)
for loss in losses:
    print(loss._value)
```
Now, the pipeline will be saturated and the overall latency will be reduced, but since all the execution requests of `infer_step` have been sent to the `MeshHostWorkers` first, and ray actors process requests in a FIFO order, all the `get_remote_buffers` requests of `loss._value` will be blocked until all the `infer_step`s finish execution, which means the user can get the output of the first `infer_step` call only after the last `infer_step` call finishes execution. 
The figure below illustrates this:
image

There are two solutions to this problem:
Solution 1: Add a flag argument in `get_remote_buffers` to enable it to return the ray_remote_ref immediately instead of blocking until the buffers are ready. Then during the initialization of `DistributedArray`, call `get_remote_buffers` and save ray_remote_ref in the class.  Since the output DistributedArray (i.e. `loss` in the above code) is constructed immediately after the `infer_step` returns, now the `get_remote_buffers` requests will be sent before the next `infer_step`'s execution request. So when the user tries to get the output of the first `infer_step` call, the value will be returned as soon as the first `infer_step` finishes execution and will not be blocked by the successive `infer_step` calls.
The figure below illustrates this:
image
Pros: This solution is completely transparent to the users.
Cons: Add additional overhead for the construction of `DistributedArray`. As noted by @zhuohan123 , if the user program produces many `DistributedArray`, the overhead may be unbearable.

Solution 2: Add an explicit function call to get the ray_remote_ref of `get_remote_buffers` instead of implicitly getting them during the initialization of `DistributedArray`. Now the program looks like this:
 ```
losses = [ ]
for batch in batches:
    loss = infer_step(params, batch)  # return a DistributedArray
    loss.get_remote_buffers_async() # add this extra line of code
    losses.append(loss)
for loss in losses:
    print(loss._value)
```
Pros:  (1) Eliminate unnecessary overhead. (2) Support batching
Cons: Not transparent to the users.

Benchmark: Solution 1 incurs about 50ms overhead per iteration, while Solution 2's overhead is negligible.

This PR implements Solution 1, while Solution 2 is in my local repository.

### CODE REVIEW

**Review by zhuohan123 — APPROVED** [review]
LGTM. Updated the code to add some todos in the future.

File: benchmark/alpa/benchmark_3d_infer_one_case_gpt_bert.py [hunk 1]
```diff
@@ -199,12 +213,26 @@ def benchmark_gpt_bert_internal(model_type,
...
-        _ = infer_step(params, batch, rngkey)
+        loss = infer_step(params, batch, rngkey)
+        loss.get_remote_buffers_async()  # unstable API
+        losses.append(loss)
+        if not stream_mode:
+            executable.sync()
+    for i, loss in enumerate(losses):
+        print(loss._value)
+        times[f"iter{i}"] = time.time() - tic
+    if stream_mode:
```
zhuohan123, line 229: cc @zhisbug 

With this PR, you can implement ORCA's iteration-level schedule and pipeline parallel schedule following this code snippet.

Issue #550, opened 2022-06-28
[FEATURE] Auto-Parallelization for Graphs with Predesignated Parameter Shardings


zhuohan123:
Currently, Alpa only supports auto parallelization of a graph without existing annotation. However, there are cases where the shardings of some parameters are provided in advance. For example, when performing inference on variable length inputs, the graphs for different should share the same parameter sharding to be able to serve these graphs concurrently efficiently. This doc outlines the steps towards fully supporting auto-parallelization with user provided shardings:

### Manual inter-operator parallelism and sharding-propagation based intra-operator parallelism

This should be our first step and should be the easiest to be implemented. We can follow the logic of [compile_create_state_executable](https://github.com/alpa-projects/alpa/blob/8d69815db9617a9e720927c39ad5bde01fd06f0f/alpa/create_state_parallel.py#L69) to:
1. Get the sharding spec and the mesh assignment for the parameters. https://github.com/alpa-projects/alpa/blob/8d69815db9617a9e720927c39ad5bde01fd06f0f/alpa/create_state_parallel.py#L84-L86
2. Slice the pipeline stages according to the user annotation. Assume the user provided stage boundaries align with the provided mesh assignment for the parameters.
3. Set the **input sharding** and run the sharding propagation pass. In the example below, we run sharding propagation with the specific output sharding: https://github.com/alpa-projects/alpa/blob/8d69815db9617a9e720927c39ad5bde01fd06f0f/alpa/create_state_parallel.py#L128-L132

### Manual inter-operator parallelism and Alpa intra-operator parallelism

In this step, we need to modify the auto-sharding ILP in Alpa to support user provided annotation. The basic idea is to force some variables in the ILP to select the user provided choice. This part has already been implemented within Google's internal auto-sharding pass. We are in the process of open-sourcing that part into the official tensoflow/XLA codebase, which should be out within two weeks. I will update this part after the code is open-sourced.

### Alpa inter- and intra-operator parallelism

In this step, we need to modify the stage-slicing DP algorithm in Alpa as follows:
1. Support running the DP algorithm with a user provided mesh slicing. In the current algorithm, we assume the user only provides a huge 2D device mesh and we will slice the mesh and the stages together.
2. Let the DP algorithm consider the parameters' mesh assignment when slicing stages. 
3. Let the layer clustering algorithm also consider parameters' mesh assignment.

I can follow up with a more detailed design for the algorithmic changes here.

PR #620, opened 2022-07-20
[Feature] Support input sharding specs assignment

comaniac:
A part of #550: Manual inter-operator parallelism and sharding-propagation based intra-operator parallelism

Improve `PipeshardParallel` to optionally accept `input_sharding`. When given, the input of each HLO module will be sharded accordingly.

See unit test for details.

cc @merrymercy @zhuohan123

### DIFF
--- a/alpa/parallel_method.py
+++ b/alpa/parallel_method.py
@@ -17,6 +17,7 @@ from typing import Callable, Optional, Sequence, Union, Any
 
 from jax import linear_util as lu
 from jax.core import AbstractValue
+from jax.interpreters import pxla
 from jax.tree_util import PyTreeDef
 import numpy as np
 
@@ -166,6 +167,7 @@ class PipeshardParallel(ParallelMethod):
         stage_option: Options of grouping layers into pipeline stages.
           Possible choices are {"uniform", "auto", alpa.AutoStageOption,
                                  alpa.ManualStageOption}
+        TBA
     """
 
     def __init__(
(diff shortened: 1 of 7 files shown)


tests/pipeline_parallel/test_set_input_shard.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def setUp(self):
+        init(cluster="ray")
+
+    def run_set_input_shard_spec(self):
+        hidden_size = 64
+
+        rngkey = jax.random.PRNGKey(0)
+
+        # Make a MLP model with 2 pipeline stages.
```
comaniac, line 19:
```suggestion
        # Make a MLP model with 2 pipeline stages
```

alpa/parallel_method.py
```diff
@@ -166,6 +167,8 @@ class PipeshardParallel(ParallelMethod):
         stage_option: Options of grouping layers into pipeline stages.
           Possible choices are {"uniform", "auto", alpa.AutoStageOption,
                                  alpa.ManualStageOption}
+        input_sharding: Options of input sharding specs for each stage.
```
merrymercy, line 170:
```suggestion
        stage_input_sharing: Options of input sharding specs for each stage.
```

Because this is not the sharding of global vars, let us use `stage_input_sharding` to avoid confusion.

merrymercy (changes requested):
Only a minor comment

alpa/parallel_method.py
comaniac: Good point. Renamed.

merrymercy (approved):

PR #632, opened 2022-08-01
Speed up DP

makslevental:
Speed up inter-op DP by moving `stage_cost <= max_stage_cost` check out of inner loop (and a few other small things). 

**TODO**: add test to show equivalence. 

I have a [short gist](https://gist.github.com/makslevental/efc065d3803abe4fa3fd171d204441a8) that verifies numerical accuracy of this refactoring (against alpa's current implementation) along with the speed improvements but I've made changes since publishing it (the gist). Will update the gist soon (and publish as a colab).

cc @zhuohan123


zhuohan123:
Thanks for the contribution! Yes, please add a correctness test to make sure the old and new DP are numerically the same.

makslevental:
> Thanks for the contribution! Yes, please add a correctness test to make sure the old and new DP are numerically the same.

Easiest place to put the correctness check was here (it's pretty tough to get alpa compiled!). So I put back the current implementation and renamed my implementations as `*_2`. You can run `test_old_dp_vs_new_dp.py` and see the correctness results and relative perf (I added all the extra `njit` keywords to your impl too just so that it's a fair comparison).

alpa/pipeline_parallel/stage_construction.py
```diff
@@ -82,7 +82,130 @@ def get_last_dp_result():
...
+        (num_layers + 1, num_layers + 1, num_devices + 1), np.inf, dtype=np.float32
+    )
+    f_stage_max = np.full(
+        (num_layers + 1, num_layers + 1, num_devices + 1), 0.0, dtype=np.float32
+    )
+    f_argmin = np.full(
+        (num_layers + 1, num_layers + 1, num_devices + 1, 3), -1, dtype=np.int32
+    )
+    f[0, num_layers, 0] = 0
+    for d in maybe_prange(1, num_devices + 1):
```
zhuohan123, line 120: There are dependencies between the iterations on different values of `d`. Can we actually use `prange` here? Even if the result is correct, can `prange` here actually accelerate the execution?

zhuohan123, line 120: There are dependencies between the iterations on different values of `d`. Can we actually use `prange` here? Even if the result is correct, can `prange` here actually accelerate the execution?

alpa/pipeline_parallel/stage_construction.py
```diff
@@ -82,7 +82,130 @@ def get_last_dp_result():
...
+    all_possible_stage_costs = np.sort(np.unique(compute_cost))
+    best_cost = np.inf
+    best_solution = None
+    last_max_stage_cost = 0.0
+    # FIXME(zhuohan): Set this gap as a tunable parameter in global config
+    gap = 1e-6
+    assert len(
+        all_possible_stage_costs), "no solution in auto stage construction."
+
+    submesh_sizes: list = MaybeNumbaList()
```
zhuohan123, line 155: Why don't simply use an `np.array` here?

zhuohan123 (commented):
The code looks good in general! Left some comments. In addition, can you also run the formatter to pass the lint checks?

alpa/pipeline_parallel/stage_construction.py
makslevental: you know you're right - there's a clear loop carried dependency (duh) so this shouldn't give numerically correct answer but it does?

>Even if the result is correct, can prange here actually accelerate the execution?

if the question is whether `prange` is actually faster, the answer is yes i think - at the minimum it is actually running a bunch of processes in parallel (you can check `htop`). whether there's some kind of synchronization cost that negates the benefit of the parallelization i don't know (i didn't carefully profile the difference with/without). either way it's not mathematically correct so i'll take it out, but it might be worth thinking about whether the inherent structure of the optimization somehow breaks the dependence (and that's why it's not coming up incorrect).

alpa/pipeline_parallel/stage_construction.py
makslevental: yea that works too (`List` didn't work because numba couldn't infer the element types or something like that).

makslevental: actually i take it back, making the replacement fails type inference too:

```
No implementation of function Function(<built-in function getitem>) found for signature:
 
 >>> getitem(array(float32, 3d, C), Tuple(int64, int64, float64))
 
There are 22 candidate implementations:
   - Of which 20 did not match due to:
   Overload of function 'getitem': File: <numerous>: Line N/A.
     With argument(s): '(array(float32, 3d, C), Tuple(int64, int64, float64))':
    No match.
   - Of which 2 did not match due to:
   Overload in function 'GetItemBuffer.generic': File: numba/core/typing/arraydecl.py: Line 166.
     With argument(s): '(array(float32, 3d, C), Tuple(int64, int64, float64))':
    Rejected as the implementation raised a specific error:
      NumbaTypeError: unsupported array index type float64 in Tuple(int64, int64, float64)
  raised from /home/mlevental/miniconda3/envs/alpa/lib/python3.9/site-packages/numba/core/typing/arraydecl.py:72

During: typing of intrinsic-call at /home/mlevental/dev_projects/alpa/alpa/pipeline_parallel/stage_construction.py (129)

File "../../alpa/pipeline_parallel/stage_construction.py", line 129:
def dp_impl_2(
    <source elided>

                    new_cost = f[s - 1, i + 1, d - n_submesh_devices] + stage_cost
                    ^
```

alpa/pipeline_parallel/stage_construction.py
```diff
@@ -159,9 +159,8 @@ def dp_2(
     assert len(
         all_possible_stage_costs), "no solution in auto stage construction."
 
-    submesh_sizes = np.empty((len(submesh_choices),), dtype=np.int64)
-    for n, m in submesh_choices:
-        submesh_sizes.append(n * m)
+    submesh_sizes = np.array([n * m for (n, m) in submesh_choices],
+                             dtype=np.int64)
```
makslevental, line 163: 👍 will remember this one

zhuohan123 (approved):
I'll merge the PR now and optimize it in future PRs.

PR #639, opened 2022-08-06
[Example] Fine-tune pre-trained OPT models from huggingface

merrymercy:
- Add an OPT fine-tuning example


zhisbug:
@merrymercy We should write a very detailed tutorial for this example once it is done?

Also, I am thinking that we should update our project README now to highlight examples (and advantages) as we have at least 4 complete tutorials:
- get started with alpa
- OPT serving
- finetuning
- Paper related benchamrking and performance tuning?

PR #647, opened 2022-08-10
Use relative path to access completions URL

aurickq:
We ran into some trouble hosting the OPT demo behind a reverse proxy like NGINX that adds a prefix to the OPT serving URLs. This PR changes the completions request to use a relative URL so the OPT service can work under and reverse proxy.


JiahaoYao:
@zhisbug btw, does the master break due to this?

PR #615, opened 2022-07-19
[FEATURE] beam size and num return sequences for interactive hosting

vinlnx:

PR #653, opened 2022-08-15
add logprobs endpoint with caching past key values

yangkevin2:
Hi, 

I'd like to be able to modify the logits before sampling at each step of generation, so I implemented a logprobs endpoint which returns logprobs based on top-p and top-k and also caches past_key_values. To avoid mixing work items with the existing completions endpoint, I made a separate queue for the new endpoint. The past_key_values cache removes old items after 10 minutes or after it gets too full. 

I was hoping to integrate the new endpoint more with the existing completions endpoint which calls huggingface's generate() method, but it seems like this might not be possible at the moment due to generate() not supporting returning past key values pending this PR https://github.com/huggingface/transformers/pull/17574. 

Also, the new endpoint does everything with batch size 1 to avoid needing to mix past_key_values from different calls, but then pads up to batch size MAX_BS=16 anyway since that's what the model interface seems to expect. Not sure if this would be a problem down the line. 

Please let me know if there's something implemented in a problematic way and I can try to fix it. Would really like to be able to run controlled generation approaches with these larger models.

### DIFF
--- a/examples/opt_serving/generator.py
+++ b/examples/opt_serving/generator.py
@@ -80,6 +80,35 @@ class GeneratorInterface:
         else:
             self.num_gpus = 1
         logger.info(f"Loading model time: {load_time:.2f}")
+    
+    def __call__(
+        self,
+        inputs,
+        cache_ids,
+        pasts=None,
+    ):
+        assert len(inputs) == 1, "enforcing batch size 1 for now to avoid messy logic with caching past key values"
+        inputs = pad_batch(inputs, self.tokenizer.pad_token_id, MAX_BS)
+        tokens = [torch.LongTensor(t) for t in inputs]
+        lengths = [len(t) for t in inputs]
+        batches = self.get_batch_iterator(
+            dataset=self.build_dataset_for_inference(tokens, lengths),
+            max_tokens=None,
+            max_sentences=MAX_BS,
+            max_positions=None,
+            ignore_invalid_inputs=False,
+        ).next_epoch_itr(shuffle=False)
+        assert len(batches) == 1, "we have enforce inner bs = 1."
+        outputs = []
+        for idx, (batch, cache_id) in enumerate(zip(batches, cache_ids)):
+            if self.torch_device == "cuda":
+                input_ids = move_to_cuda(batch['src_tokens'])[:, :-1]
+            else:
+                input_ids = batch['src_tokens'][:, :-1]
+            attention_mask = self.model_wrapper._prepare_attention_mask_for_generation(input_ids, pad_token_id=self.model_wrapper.config.pad_token_id, eos_token_id=self.model_wrapper.config.eos_token_id)
+            model_inputs = self.model_wrapper.prepare_inputs_for_generation(input_ids, past=pasts[cache_id][1] if pasts is not None else None, attention_mask=attention_mask)
+            outputs.append(self.model_wrapper(**model_inputs))
+        return outputs
 
     def generate(
         self,
--- a/examples/opt_serving/interactive_hosted.py
+++ b/examples/opt_serving/interactive_hosted.py
@@ -6,6 +6,9 @@ import threading
 import os
 import logging.handlers
 import traceback
+from collections import defaultdict
+import uuid
+import time
 
 import random
 import torch
@@ -18,14 +21,17 @@ from opt_serving.service.responses import OAIResponse
 from opt_serving.service.utils import encode_fn, build_logger
 from opt_serving.service.workers import WorkItem
 from opt_serving.service.constants import MAX_SEQ_LEN, MAX_BATCH_TOKENS, TIMEOUT_MS, \
-    MAX_BS, NUM_BEAMS, NUM_RETURN_SEQ
+    MAX_BS, NUM_BEAMS, NUM_RETURN_SEQ, LOGPROBS_PAST_CACHE_SIZE_LIMIT, LOGPROBS_PAST_CACHE_TIMEOUT
 
 app = Flask(__name__, template_folder='service')
 
 # The global text generator
 generator: GeneratorInterface = None
-# The request queue
-batch_queue = PriorityQueueRingShard()
+# The request queues
+generate_batch_queue = PriorityQueueRingShard()
+logprobs_batch_queue = PriorityQueueRingShard()
+# Past key caches
+logprobs_past_cache = defaultdict(lambda: (0, None))
 # Logging
 logger = build_logger()
 
@@ -41,7 +47,7 @@ def batching_loop(timeout=TIMEOUT_MS, max_tokens=MAX_BATCH_TOKENS, max_bs=MAX_BS
     batching_loop is an infinite loop responsible for executing generations.
 
     GPUs benefit from batching requests, but we expect workloads to come
-    in non-uniformly. This loop groups requests together (via batch_queue)
+    in non-uniformly. This loop groups requests together (via generate_batch_queue)
     and executes them in one batch. In order to keep latency low, unfilled
     batches are executed within a window of :timeout: milliseconds.
 
@@ -60,13 +66,20 @@ def batching_loop(timeout=TIMEOUT_MS, max_tokens=MAX_BATCH_TOKENS, max_bs=MAX_BS
     # TODO(roller):
     # - group by generation type, topp etc, as we cannot share these
     # - modify timeout logic to be cumulative
-    batch = []
+    generate_batch = []
+    logprobs_batch = []
     while True:
-        try:
-            # for now, we only have 1 worker, so can always index to shard 0
-            target_queue = batch_queue.queue_shards[0].get_largest_queue()
-            if not target_queue:
-                continue
+        generate_batch, made_model_call = generate_loop(generate_batch, timeout=timeout, max_tokens=max_tokens, max_bs=max_bs)
+        if not made_model_call: # only process logprobs requests if we're not actively serving generate requests
+            logprobs_batch, _ = logprobs_loop(logprobs_batch, timeout=timeout, max_tokens=max_tokens, max_bs=max_bs)
+
+
+def generate_loop(generate_batch, timeout=TIMEOUT_MS, max_tokens=MAX_BATCH_TOKENS, max_bs=MAX_BS):
+    made_model_call = False
+    try: # generate endpoint
+        # for now, we only have 1 worker, so can always index to shard 0
+        target_queue = generate_batch_queue.queue_shards[0].get_largest_queue()
+        if target_queue:
             # dynamic batching: group like-sized items to reduce the cost
             # of padding. See PR#20 for additional context.
             item = target_queue.get(timeout=timeout / 1000)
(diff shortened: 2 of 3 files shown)


examples/opt_serving/interactive_hosted.py
```diff
@@ -235,6 +313,63 @@ def completions(engine=None):
     return OAIResponse(results).__dict__()
 
 
+@app.route("/logprobs", methods=["POST"])
```
zhuohan123, line 316: Can we have a flag to turn on/off this API when starting the interactive host? And also, can we set a password or security key so that we can only allow private access to this API?

examples/opt_serving/interactive_hosted.py
```diff
@@ -105,13 +116,80 @@ def batching_loop(timeout=TIMEOUT_MS, max_tokens=MAX_BATCH_TOKENS, max_bs=MAX_BS
...
+                    for work_item, gen in zip(logprobs_batch, generations):
+                        work_item.return_queue.put((work_item.uid, gen))
+
+                    logprobs_batch.clear()
+        
+        # clear old keys in PAST_CACHE which haven't been used recently
+        existing_cache_ids = list(PAST_CACHE.keys())
+        for cache_id in existing_cache_ids:
+            if time.time() - PAST_CACHE[cache_id][0] > PAST_CACHE_TIMEOUT:
+                del PAST_CACHE[cache_id]
```
zhuohan123, line 191: If I understand correctly, PAST_CACHE will be a huge cache that occupies a large portion of GPU memory. If we only use the TIMEOUT mechanism, will it make our original service OOM due to the cache in GPU?

zhuohan123 (changes requested):
Left some review comments. @zhisbug please also review this PR since it affects our main service.

examples/opt_serving/generator.py
```diff
@@ -137,6 +137,50 @@ def load_model(self):
...
+        for new_input in new_inputs:
+            ori_len = len(new_input)
+            if len(new_input) < max_len:
+                new_input.extend([pad_value for _ in range(max_len - ori_len)])
+        # pad to bs = MAX_BS
+        if bs < MAX_BS:
+            new_inputs.extend([[pad_value for _ in range(max_len)] for _ in range(MAX_BS - bs)])
+        return new_inputs
+    
+    def __call__(
```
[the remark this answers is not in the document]
yangkevin2: I used `__call__` because it's essentially just wrapping the OPT model's forward function, which would normally be accessed by just doing e.g. `model(**args)`. I could rename it to `forward` instead and explicitly call `.forward()` if that would be preferred? What do you think?

examples/opt_serving/generator.py
```diff
@@ -137,6 +137,50 @@ def load_model(self):
...
+        self,
+        inputs,
+        cache_ids,
+        pasts=None,
+    ):
+        assert len(inputs) == 1, "enforcing batch size 1 for now to avoid messy logic with caching past key values"
```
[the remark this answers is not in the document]
yangkevin2: I plan to call it frequently, but only one request (batch size 1) at a time since the requests are sequential in nature, so I don't think the batching would make things faster for my use case. So I didn't implement it for now. Is that fine for the current PR?

examples/opt_serving/interactive_hosted.py
```diff
@@ -105,13 +116,80 @@ def batching_loop(timeout=TIMEOUT_MS, max_tokens=MAX_BATCH_TOKENS, max_bs=MAX_BS
...
+                for work_item, gen in zip(generate_batch, generations):
                     work_item.return_queue.put((work_item.uid, gen))
 
-                batch.clear()
-            else:
-                # back to the loop
-                continue
+                generate_batch.clear()
+
+        try: # logprobs endpoint
```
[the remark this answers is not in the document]
yangkevin2: Sure, will refactor

examples/opt_serving/interactive_hosted.py
yangkevin2: Yeah I can definitely look into adding a flag and have it off by default or something. I'm not sure what's the best practice for the password/security key though, is there some method you have in mind?

examples/opt_serving/interactive_hosted.py
yangkevin2: I also wrote a mechanism to remove the oldest item once it exceeds a specified maximum size, regardless of timeout, based on `PAST_CACHE_SIZE_LIMIT`. So the cache should be limited to some maximum size. 

Additionally, it might be possible to cache only the relevant keys instead of the full padded batch of size 16 by manipulating the past_key_values, but I wasn't sure if constantly reconstructing the padded keys on GPU would introduce a lot of overhead so I didn't do it for now. 

I'm not sure how much of a limitation the memory would be on your server; happy to modify the numbers/implementation based on what you think the right tradeoff is.

zhisbug:
@yangkevin2 Thanks for the great work -- I'll review by tonight

examples/opt_serving/interactive_hosted.py
```diff
@@ -49,15 +56,26 @@ def batching_loop(timeout=TIMEOUT_MS, max_tokens=MAX_BATCH_TOKENS, max_bs=MAX_BS
...
-    batch = []
+    generate_batch = []
+    logprobs_batch = []
     while True:
-        try:
-            # for now, we only have 1 worker, so can always index to shard 0
-            target_queue = BATCH_QUEUE.queue_shards[0].get_largest_queue()
-            if not target_queue:
-                continue
+        generate_batch, made_model_call = generate_loop(generate_batch, timeout=timeout, max_tokens=max_tokens, max_bs=max_bs)
```
zhisbug, line 68: I suppose this means if the external users are submitting requests, the returned `made_model_call` of their requests will be True. So your endpoint won't run?

But if at any point, you are submitting requests while any other users are not submitting requests (hence the old batching queue is empty), your requests will go through into the `if not made_model_call` branch, which will then go to the function you added in `generator.py`? 

Please correct if some of my understanding is something wrong.

examples/opt_serving/interactive_hosted.py
```diff
@@ -247,6 +397,12 @@ def index():
     parser.add_argument("--model", type=str, default="alpa/opt-125m")
     parser.add_argument("--path", type=str, default="/home/ubuntu/opt_weights/")
     parser.add_argument("--port", type=int, default=DEFAULT_PORT)
+    parser.add_argument("--serve-logprobs", action="store_true", default=False)
     args = parser.parse_args()
 
+    if args.serve_logprobs:
+        logprobs.provide_automatic_options = False
+        logprobs.methods = ['POST']
```
zhisbug, line 405: How do you plan to query the endpoint? Do you have client code? Would also be good to share.

Also, we can setup this on AWS and use your client code to do a few tests.

After things are tested stable, we can discuss a time window (say, some weekend 1 - 5am) when your program can run with full capacity

examples/opt_serving/interactive_hosted.py
```diff
@@ -67,51 +85,126 @@ def batching_loop(timeout=TIMEOUT_MS, max_tokens=MAX_BATCH_TOKENS, max_bs=MAX_BS
...
+                    if key in ro:
+                        request_object[key] = ro[key]
                 # do the actual generations
                 request_object["seed"] = random.randint(1, 20000)
-                generations = generator.generate(**request_object)
+                outputs = generator(request_object['inputs'], cache_ids, pasts=PAST_CACHE)
+                generations = []
+                for cache_id, output in zip(cache_ids, outputs):
+                    PAST_CACHE[cache_id] = (time.time(), output.past_key_values) # add to or update the cache with newly computed values
+                    if len(PAST_CACHE) > PAST_CACHE_SIZE_LIMIT: # delete oldest key in cache if cache too big
```
zhisbug, line 185: What does this cache mean? is it the attention cache or something else?

zhisbug (changes requested):

zhisbug:
@yangkevin2 please take a look at my comments! thanks you!

yangkevin2:
Thanks for the detailed feedback! I'm out traveling this weekend but will make changes accordingly on Monday.

examples/opt_serving/generator.py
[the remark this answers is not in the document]
merrymercy: SGTM

examples/opt_serving/generator.py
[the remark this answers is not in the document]
merrymercy: If you only submit one request at a time, the speed will be very slow!
1. You have to pay the network overhead for each request. But with batching, you can amortize the overhead
2. The latency of batch-size =1 and batch-size = 16 is the same. By batching, you can get 16x better throughput.

It is fine for the current PR because it is a separate endpoint. Later, if you find the generation speed too slow, you may apply batching

examples/opt_serving/generator.py
```diff
@@ -137,6 +137,50 @@ def load_model(self):
...
+            dataset=self.build_dataset_for_inference(tokens, lengths),
+            max_tokens=None,
+            max_sentences=MAX_BS,
+            max_positions=None,
+            ignore_invalid_inputs=False,
+        ).next_epoch_itr(shuffle=False)
+        assert len(batches) == 1, "we have enforce inner bs = 1."
+        outputs = []
+        for idx, (batch, cache_id) in enumerate(zip(batches, cache_ids)):
+            input_ids = move_to_cuda(batch['src_tokens'])[:, :-1]
```
merrymercy, line 178: We change the default torch device to CPU (while alpa still runs on GPU) to avoid the memory conflicts between torch and alpa.

Please update accordingly
https://github.com/alpa-projects/alpa/blob/6085a0877a999d60cfc4409e25ef51cd96703eea/examples/opt_serving/generator.py#L171-L172

merrymercy (commented):
It looks good to me generally. During last few days, we made some big improvements, such as beam search (https://github.com/alpa-projects/alpa/pull/615), faster generation (https://github.com/alpa-projects/alpa/pull/661), and code clean-up (https://github.com/alpa-projects/alpa/pull/667). Please rebase!

examples/opt_serving/interactive_hosted.py
```diff
@@ -20,7 +23,11 @@
 from opt_serving.service.constants import MAX_SEQ_LEN, MAX_BATCH_TOKENS, DEFAULT_PORT, TIMEOUT_MS, MAX_BS
 
 app = Flask(__name__)
-BATCH_QUEUE = PriorityQueueRingShard()
+GENERATE_BATCH_QUEUE = PriorityQueueRingShard()
+LOGPROBS_BATCH_QUEUE = PriorityQueueRingShard()
+PAST_CACHE = defaultdict(lambda: (0, None))
```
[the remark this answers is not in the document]
yangkevin2: Done!

examples/opt_serving/interactive_hosted.py
```diff
@@ -67,51 +85,126 @@ def batching_loop(timeout=TIMEOUT_MS, max_tokens=MAX_BATCH_TOKENS, max_bs=MAX_BS
...
+                ro = work_item.data
+                request_object["inputs"].append(ro["input"])
+                request_object["min_tokens"].append(ro.get("min_tokens", 0))
+                request_object["max_tokens"].append(
+                    ro.get("max_tokens", MAX_SEQ_LEN))
+                cache_ids.append(ro.get("cache_id", None))
+                # assumption: everyone has the same remaining args
+                for key in [
+                        "temperature",
+                        "top_p",
```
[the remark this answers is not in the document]
yangkevin2: Yeah, I envisioned possibly wanting to have both so I implemented both for this endpoint (but didn't touch the completions endpoint)

examples/opt_serving/generator.py
```diff
@@ -137,6 +137,50 @@ def load_model(self):
...
+            attention_mask = self.model_wrapper._prepare_attention_mask_for_generation(input_ids, pad_token_id=self.model_wrapper.config.pad_token_id, eos_token_id=self.model_wrapper.config.eos_token_id)
+            model_inputs = self.model_wrapper.prepare_inputs_for_generation(input_ids, past=pasts[cache_id][1] if pasts is not None else None, attention_mask=attention_mask)
```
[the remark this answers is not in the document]
yangkevin2: I did notice that the current batching logic was being taken from the first sentence and being applied to everything else, but ignored this issue because in my current implementation of the new endpoint it's just forcing batch size 1. So I think it should be ok?

examples/opt_serving/interactive_hosted.py
yangkevin2: Yeah, Zhuohan suggested that I prioritize the existing endpoint, so I implemented this logic to prioritize requests to completions and only run my endpoint if the queue is empty.

examples/opt_serving/interactive_hosted.py
yangkevin2: It's a cache for the past_key_values, yeah.

examples/opt_serving/interactive_hosted.py
yangkevin2: Here's some simple example code I've been using to test:

```
import requests
import json

from scipy.special import softmax
import numpy as np
from transformers import AutoTokenizer

tokenizer = AutoTokenizer.from_pretrained('facebook/opt-125m')

url = "http://localhost:9054/logprobs"
headers = {'Content-type': 'application/json', 'Accept': 'text/plain'}
data = {
    "prompt":[int(x) for x in tokenizer.encode("mary had a little")],
    "max_tokens":"64",
    "temperature":"0.7",
    "top_p":"0.5",
    "top_k":"100",
}

r = requests.post(url, data=json.dumps(data), headers=headers)
# import pdb; pdb.set_trace()

for i in range(20):
    output = r.json()
    data['cache_id'] = output['cache_id']
    distribution = np.array([-1e8 for _ in range(tokenizer.vocab_size)])
    print(len(output['logprobs']))
    for idx, logprob in zip(output['indices'], output['logprobs']):
        distribution[idx] = logprob
    distribution = softmax(distribution)
    token = np.random.choice(np.arange(len(distribution)), p=distribution)
    data['prompt'].append(int(token))
    print(tokenizer.decode(data['prompt']))
    r = requests.post(url, data=json.dumps(data), headers=headers)

import pdb; pdb.set_trace()
```

After things are tested + stable, I was hoping to be able to query the endpoint at any time, and just give priority to the completions endpoint when there's traffic (as I implemented with the `made_model_call` logic). Would that be possible?

examples/opt_serving/generator.py
yangkevin2: Updated!

yangkevin2:
Hi, I've rebased and (I think) addressed all the existing comments, minus @zhisbug 's comments about the incorrect generation params with batching since I'm currently not batching with the new endpoint anyway.

Thinking about it more carefully, I may actually want to batch though, so I'll work on that this week and see if it's feasible (might be tricky with having to also cache past_key_values; if necessary I could force the user to pass one entire batch at a time or something, and advance the batch all together, to avoid mixing batches). 

I also noticed that when testing on AWS, it took a while longer to generate sequentially one token at a time using the new endpoint compared to just using the normal completions endpoint, even when I ran the client code from the same machine. Not sure if there's still network latency with python's requests module even when running from the same machine or if there's some other bottleneck; will dig around a bit more. 

Also, an alternative implementation that doesn't require so many network calls could be to use the `logits_warper` functionality in huggingface's generate (https://huggingface.co/docs/transformers/main_classes/text_generation#transformers.generation_utils.GenerationMixin.sample.logits_warper). This might actually be cleaner code-wise since it'd just be an extra argument to the completions endpoint, but might require being able to pass in arbitrary functions as a logits_warper and running them on the server, so might need some extra security? Curious if you guys would prefer going this route or the current one.

merrymercy:
I port your code and did some clean-up and fixes in #669 
The major fix is the batching logic.
1. Your current implementation of the priority is not correct. We should use `receive_item` instead of `made_model_call` to implement priority
2. The slow generation speed also comes from the batching logic. The default code has a `timeout=2000ms` argument, so it always waits for 2 seconds to wait for batching opportunities before actually responding to a request. Now, I set a timeout=0 for the logprob endpoint, so it is much faster.

We can go ahead with the current approach and let you start trying things. If you find the speed too slow, I think `logits_warper` might be the best way to go. We can give you a password and only support this API with a correct password, so other people cannot run arbitrary python code. If we want to implement the `logits_warper` API, the question is "Can all your logits modification code be wrapped in a stateless python function?"

merrymercy:
The logprob endpoint and a test script are added (https://github.com/alpa-projects/alpa/blob/main/examples/opt_serving/test_logprobs.py).
I tried to set it up on our server. The completion API works, but I got `403 Forbidden` error for the logprobs API. I guess we need to update some firewall settings, @zhisbug will figure this out.

yangkevin2:
Great, thank you for fixing the priority + timeout! Will keep the `logits_warper` in mind but I think it might not be necessary for now. 

Anyway, please just let me know when the main OPT-175B server is ready to serve logprobs requests, and/or if anything comes up that I need to fix.

PR #657, opened 2022-08-16
[Alpa Runtime] update the alpa api for resources awareness

JiahaoYao:
> Hello! Sorry maybe a silly question, but is there a configuration for alpa to use only a subset of the # of gpus? (Ie. only run 2 cores when 8 cores are available) Do I have to create a device mesh for it? 

People prefer to set the configuration for the alpa instead of using the whole cluster. This PR is try to enrich the resources allocation with a easy-to-use api.

image

@zhisbug @merrymercy 

- [x] adding the `cpu=1e-4` for the `daemonworker`.

### DIFF
--- a/alpa/api.py
+++ b/alpa/api.py
@@ -21,7 +21,9 @@ traceback_util.register_exclusion(__file__)
 is_initialized = False
 
 
-def init(cluster: str = "ray"):
+def init(cluster: str = "ray",
+         devices_per_node: int = None,
+         num_nodes: int = None):
     """Initialize the global environment.
 
     Args:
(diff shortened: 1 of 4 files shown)


JiahaoYao:
btw, @merrymercy do u have any suggestions about the uni-test for this pr?

JiahaoYao:
ready to review with the test

merrymercy (approved):

zhisbug:
@JiahaoYao any chance you can add a documentation for this API?

JiahaoYao:
SGTM, @zhisbug where is the best place for this documentation?

PR #564, opened 2022-06-30
[Feature] Support multiple executables in OPT serving

comaniac:
The initial support of multiple executables to cover various input prompt lengths.

Usage (note that this is only applicable to autoregressive mode):
```
python benchmark_text_gen.py --model alpa/opt-2.7b --path ... --multi-executable
```

Since this is only for benchmarking, currently we check all prompts we are going to run and make a list of decoder lengths accordingly.

Next steps:
- Enable parameter sharing between executables (#550). Now we encounter the following error, which should be caused by different input shardings.
```
  File "/home/ubuntu/alpa/alpa/pipeline_parallel/pipeshard_executable.py", line 221, in __call__
    out = self.launch_on_driver(*args_flat)
  File "/home/ubuntu/alpa/alpa/pipeline_parallel/pipeshard_executable.py", line 150, in launch_on_driver
    self.num_batch, mesh_args)
  File "/home/ubuntu/alpa/alpa/device_mesh.py", line 1117, in shard_args_to_bufs
    assert replica.indices == indices
AssertionError
```
- Fix the bug of crashing with non-power-of-2 prompt (#561). The current error is:
```
  File "/home/ubuntu/alpa/alpa/pipeline_parallel/cross_mesh_resharding.py", line 1067, in dst_tile_to_src_tiles_map
    self._dst_tile_to_src_tiles_map = self.generate_src_dst_map()
  File "/home/ubuntu/alpa/alpa/pipeline_parallel/cross_mesh_resharding.py", line 1085, in generate_src_dst_map
    self._look_up_dst_tile_from_src(tile))
  File "/home/ubuntu/alpa/alpa/pipeline_parallel/cross_mesh_resharding.py", line 1113, in _look_up_dst_tile_from_src
    assert not ragged
AssertionError
```
- Support input padding. Currently we compile one executable for a prompt. However, ideally we should limit the executable number and pad the input prompt to fit the pre-compiled executable with the nearest decoding length.
- Improve performance. Now multiple executables will introduce some overheads at the certain iteration (e.g., 50) during text generation, so that its end-to-end latency is longer than the single executable. I'm still investigating the root cause.

cc @zhisbug @zhuohan123 @merrymercy

PR #661, opened 2022-08-20
[OPT] Compile multiple executables to process prompts

merrymercy:
Compile multiple executables to process prompts. This is based on #564 and adds the support of padding and batching.

Now we can compile multiple executables with pre-defined lengths and alpa will compose these executables to process prompts.

For example, we can compile executables with lengths [1, 128, 256, 1024]. For a prompt with len=300, alpa will pad it to len=384 and use two executables (256 and 128) to process the prompt.

## Performance improvement

Single executable
```
python3 benchmark_text_gen.py --debug --model alpa/opt-2.7b --n-iter 1 --pad-to-max-length 250 --max-length 270
```
Speed:  50 token/s

Multi executable
```
python3 benchmark_text_gen.py --debug --model alpa/opt-2.7b --n-iter 1 --pad-to-max-length 250 --max-length 270 --multi
```
Speed:  570 token/s

## Notes
This is a strict performance optimization. If we only compile one executable, the performance is the same as the old one. I confirmed that all existing functionalities work and the generation results are correct.


merrymercy:
cc @zhisbug @comaniac @zhuohan123

PR #664, opened 2022-08-21
[OPT] Download weights automatically from huggingface

merrymercy:
- Now for OPT 125M-66B, we don't need to do manual weight conversion. We can automatically download the weights from hugging face and convert them to alpa format if there are no existing cached weights.
- Update weight format from `qvk_combined` to `qkv_combined`

### DIFF
--- a/examples/opt_serving/model/wrapper.py
+++ b/examples/opt_serving/model/wrapper.py
@@ -533,7 +533,10 @@ def pad_attention_mask(mask, max_target_positions):
 def download_weights(model_name, path):
     """Download weights from huggingface."""
     facebook_model_name = "facebook/" + model_name
-    print(f"Download pre-trained pytorch weights of {model_name} from huggingface ...")
+    print(f"Load the pre-trained pytorch weights of {model_name} from huggingface. "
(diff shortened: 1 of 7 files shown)


zhisbug (approved):

examples/opt_serving/model/wrapper.py
```diff
@@ -516,3 +528,22 @@ def pad_attention_mask(mask, max_target_positions):
...
+
+
+def download_weights(model_name, path):
+    """Download weights from huggingface."""
+    facebook_model_name = "facebook/" + model_name
+    print(f"Load the pre-trained pytorch weights of {model_name} from huggingface. "
+          f"The downloading and cpu loading can take dozens of minutes. "
+          f"If it seems to get stuck, you can monitor the progress by "
+          f"checking the memory usage of this process.")
+    model = OPTForCausalLM.from_pretrained(facebook_model_name, torch_dtype=torch.float16)
```
zhisbug, line 540: this one will go OOM for models > 2.7b?

merrymercy: It uses CPU, so it won't OOM

PR #672, opened 2022-08-24
[FEATURE] Split replicated equations and insert cross mesh collective communication

ZYHowell:
This PR is to optimize functions like `clip_by_global_norm`, which computes a small tensor (in this example, it's just a scalar) using relatively large tensors from multiple meshes.

Previously, when the apply gradient slicing algorithm meets this pattern, it will finally replicate these equations to all meshes that use the output scalar, then set all input tensors replicated on all these meshes. This results in an OOM behavior since all gradients are replicated on all meshes.

In this PR, such replicated eqns are detected and rewritten to split them onto different meshes. Then it inserts an all-reduce in the end to get the correct value. The all-reduce is a cross-mesh one, so this PR also adds supports for the cross-mesh collective communication.

This PR is together with https://github.com/alpa-projects/tensorflow-alpa/pull/131.


alpa/device_mesh.py
```diff
@@ -2261,6 +2273,27 @@ def get_global_num_devices():
...
+        device_strs.extend(mesh.device_strs)
+    world_size = len(workers)
+    backend = "nccl"
+    group_name = ",".join(device_strs)
+    refs = []
+    for rank, worker in enumerate(workers):
+        ref = worker.create_and_set_cross_mesh_communicators.remote(
+            world_size, rank, backend, group_name)
+        refs.append(ref)
+    ray.get(refs)
```
merrymercy, line 2294: Is this sync necessary? @ZYHowell

alpa/device_mesh.py
```diff
@@ -2261,6 +2273,27 @@ def get_global_num_devices():
     raise RuntimeError("Please call alpa.init first")
 
 
+########################################
```
merrymercy, line 2276: wrong comments

PR #679, opened 2022-08-27
limited batching support for logprobs endpoint

yangkevin2:
Hi, 

I implemented a limited version of batching for the logprobs endpoint. Specifically, the user can now send up to `MAX_BS` prompts together, each with the same number of tokens, and they'll be processed and cached together as one WorkItem. Then the user can update each of those prompts by one token and re-query with the same batch in the same order. I modified `test_logprobs.py` to show an example. 

This should be all that I need for my use case, but it currently doesn't support mixing prompts from different batches, or using prompts of different lengths in the same batch. The latter is checked via an assert; without the assert, it will run without crashing, but the result is incorrect even after I ensure the `input_ids` and `attention_mask` are correct, so I assume there's something going wrong with the `past_key_values` + padding in the presence of different-length prompts that I haven't yet looked into.

Is this fine to merge as initial batching support, and leave the rest for later?

(I also removed the top-p option, figuring the user can do their own manipulation given the logprobs, and just kept top-k in case they're network-limited).

### DIFF
--- a/examples/opt_serving/client.py
+++ b/examples/opt_serving/client.py
@@ -57,7 +57,6 @@ class Client(object):
         """TODO(yangkevin2)"""
         pload = {
             "prompt": prompt,
-            "top_p": top_p,
             "top_k": top_k,
             "redirect_logprobs": True
         }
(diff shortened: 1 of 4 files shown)


merrymercy (approved):
LGTM. I will tune some parameters (LOGPROBS_PAST_CACHE_SIZE_LIMIT) according to our cluster's GPU capacity and deploy it by the next Monday.

PR #681, opened 2022-08-28
[FEATURE] Profile only single layers (1D) instead of combination of layers (2D)

zhuohan123:
In Alpa's original auto inter-op parallel algorithm, we need to profile all possible configurations of stages, which in total has L^2 choices where L is the number of layers. In this PR, we aim to reduce this cost by only profiling L layers and calculating the stage cost by combining individual layer profiling results. To support this feature, this PR includes the following modifications:

1. To easily combine different layer results, we did a major refactor to profile the forward and backward pass separately. Specifically, during stage profiling, we call each pass a "module" (e.g., forward module, backward module) and each stage will have multiple modules. We profile each module individually.
2. We add logic to combine the results of 1D layers for a 2D stage.
3. Rename multiple variables and function names for easier understanding.


ZYHowell (commented):
Just wanna make sure my conclusion is correct:
1. Files other than `apply_grad.py` and `stage_profiling.py` are only modified for API name change or delete unused args and outputs;
2. `apply_grad.py` is modified only by moving the following functions: a) filter `DropVar` in the microbatch boundary; b) `apply_grad_get_mean`; c) `gen_dependency_with_stages`;
3. `stage_profiling.py`: profile only forward and backward of each stage(or layer) separately by: 
  a) in `generate_stage_info`, mark and collect info of forward, backward and apply_grad separately. The return value is `StageConfig` which includes a `CompileConfig` for the merged module and a `ProfileConfig` for each profiled module;
  b) in `CompileWorker.compile`, the output `CompileOutput` stores different parts in different `acc_grad_modules`, and each part is stored as a `ModuleCompileOutput`;
  c) in `profile_all`, each part is submitted to the worker pool separately, but they(fwd and bwd) are all stored in the same `ProfileResult`.
The `stage_profiling.py` also support profiling each layer and sum up to approximate for each stage configuration. The compute cost is directly added, and the memory estimation is handled in `get_max_n_succ_stages`.

alpa/pipeline_parallel/stage_profiling.py
```diff
@@ -493,115 +570,480 @@ def profile_all(stages, compiled_outputs: Sequence[CompileOutput], meshes,
...
+    assert n_modules == 2, "Only support forward and backward module"
+    assert all(result.n_modules == n_modules for result in profile_results)
+
+    stage_no = n_modules * len(profile_results)
+    # eliminate_time[var] = k means that the variable can be eliminated after
+    # stage k.
+    eliminate_time = {}
+    required_outvars = set()
+    for i in reversed(range(n_modules)):
+        for stage_result in reversed(profile_results):
```
ZYHowell, line 688: In the backward, you do not need to reverse it because the execution order is reversed. The same applies to the next loop that you need to explicitly reverse the backward part there.

alpa/pipeline_parallel/stage_profiling.py
```diff
@@ -493,115 +570,480 @@ def profile_all(stages, compiled_outputs: Sequence[CompileOutput], meshes,
...
+            var_to_be_eliminated = []
+            for var in env:
+                if (var not in required_outvars and
+                    (var not in eliminate_time or
+                     eliminate_time[var] >= stage_no)):
+                    var_to_be_eliminated.append(var)
+            for var in var_to_be_eliminated:
+                del env[var]
+            stage_no += 1
+        intermediate_size = sum(env.values())
```
ZYHowell, line 729: why do we compute that in each iteration of the loop, but use it outside of the loop?

ZYHowell (changes requested):

alpa/pipeline_parallel/stage_profiling.py
zhuohan123: Here I assume `n_modules == 2` and the `intermediate_size` captures the variables that are from the forward pass that is used in the backward pass.

ZYHowell: In case of that, doesn't it compute `intermediate_size` only when `i==0`?

alpa/pipeline_parallel/stage_profiling.py
```diff
@@ -145,12 +194,11 @@ def compile_stage_for_profiling(self, stage_id, config: CompileConfig,
             proto: The proto of compiled executable
             stage_plan: The sharding strategy from auto sharding
         """
-        self.cnt += 1
 
         # Compile with search to get sharding annotations.
         other_kwargs = {
             "logical_mesh": logical_mesh,
-            "return_mode": "stages_and_hook",
+            "return_mode": "stages",
```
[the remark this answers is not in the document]
zhuohan123: Will remove this mode in a follow up PR.

alpa/pipeline_parallel/stage_profiling.py
zhuohan123: oh, you are right! thanks for catching this!

alpa/pipeline_parallel/stage_profiling.py
zhuohan123: That's a good point! Thanks for catching this.

PR #718, opened 2022-09-26
[Docs] Update the the doc to use same snippet between jax & alpa train_step

jiaodong:
Docstring says 

```
The body of this function is the same as the ``train_step`` above. The only difference is to decorate it with ``alpa.paralellize``.
```

But it's not quite the case in docs. We should have the right drop-in replacement impression for user.


zhuohan123 (approved):

PR #727, opened 2022-10-02
Fix bloom model single-device multi-batch input output mismatch

ddxxdd-code:
Modify attention mask in wrapper to always exclude padding tokens being included in attention to match multi-batch output with single batch input.


examples/llm_serving/model/bloom_model.py
```diff
@@ -31,7 +31,7 @@
 from tqdm import tqdm
 
 from llm_serving.model.opt_model import (init_cache_aval, init_mask_aval,
-    init_cache_np, init_multi_executable_cache_dis_array)
+    init_cache_np, init_cache_dis_array, init_multi_executable_cache_dis_array)
```
merrymercy, line 34: Remove this useless import

examples/llm_serving/model/bloom_model.py
```diff
@@ -253,7 +255,7 @@ def __call__(self, x):
 
 class FlaxBloomMLP(nn.Module):
     config: BloomConfig
-    dtype: jnp.dtype = jnp.float16
+    dtype: jnp.dtype = jnp.float32
```
merrymercy, line 258: Why are all the dtype modifications needed in this file? Please explain. Otherwise, please revert them.

examples/llm_serving/model/wrapper.py
```diff
@@ -6,13 +6,10 @@
 import warnings
 
 import alpa
+from alpa import timers
```
merrymercy, line 9: Rebase and revert all debugging changes. The lastest version fixed the import problem with opt_1d.

merrymercy (changes requested):
All you need to do is just update this value to 3 for bloom models
https://github.com/alpa-projects/alpa/blob/e932d3814730bad171377555f8855d0c636f171e/examples/llm_serving/model/wrapper.py#L47
Also, please update other token ids (bos_token_id, eos_token_id).

examples/llm_serving/model/bloom_model.py
ddxxdd-code: These are changed to maximally match the hf flax bloom model, and to match the logits on each step to the same accuracy. I'll try to change these to fp16 and check if the output changes.

examples/llm_serving/model/bloom_model.py
ddxxdd-code: Resolved, in the latest commits.

examples/llm_serving/model/bloom_model.py
ddxxdd-code: The one in FlaxBloomMLP has been changed to float16.
The one in FlaxBloomForCausalLMModule.setup()'s lm_head is kept.
Below are two runs, top one with FlaxBloomForCausalLMModule.setup()'s lm_head set to float16 and the bottom one is float32:
Screen Shot 2022-10-03 at 09 51 16
The outputs are different, so this float32 to be kept.

examples/llm_serving/model/wrapper.py
ddxxdd-code: Finished in the latest commits. I add the change of pad_token_id in a if statement when model is bloom.

merrymercy:
Please use `git rebase` instead of `git pull` or `git merge` to keep your history clean. Your current change history in this page (https://github.com/alpa-projects/alpa/pull/727/files) is messed up

ddxxdd-code:
Just merged.

merrymercy (approved):

examples/llm_serving/model/wrapper.py
```diff
@@ -591,6 +590,8 @@ def run_one(_executable, _input_ids, _past_key_values, _attention_mask, num_inte
                                    output.hidden_states, output.attentions)
 
     inference_func_config = InferenceFuncConfig()
+    if "bloom" in model_name:
+        inference_func_config.pad_token_id = 3
```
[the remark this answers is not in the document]
ddxxdd-code: eos_token_id is correct. No bos_token_id dpecified in BloomConfig

PR #746, opened 2022-10-14
[EXAMPLE] port stable diffusion model into alpa

A1trialist:
Added Stable-Diffusion(JAX) to examples/ 
Successfully run on a 2-GPU single node, but with a performance worse than pmap. (Currently using alpa.DataParallel()). 
Cleaned-up version.

### DIFF
--- a/examples/stable-diffusion-jax/README.md
+++ b/examples/stable-diffusion-jax/README.md
@@ -1,0 +1,170 @@
+# Setup
+
+## PyTorch Version
+
+GitHub: https://github.com/CompVis/stable-diffusion
+
+Paper: [High-Resolution Image Synthesis With Latent Diffusion Models](https://arxiv.org/abs/2112.10752)
+
+### Notes
+
+1. Clone the repository and setup
+
+   `git clone https://github.com/CompVis/stable-diffusion.git`
+
+   `cd stable-diffusion`
+
+   `conda env create -f environment.yaml`
+
+   `conda activate ldm`
+
+   **Your pytorch version and CUDA version should match here!**
+
+   You may use 
+
+   ```python
+   import torch
+   torch.ones((2,3)).cuda()
+   ```
+   to check.
+
+2. (Alternative)
+
+   You may update the environment of [latent diffusion](https://github.com/CompVis/latent-diffusion) by running
+
+   ```bash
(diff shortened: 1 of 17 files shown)


examples/stable-diffusion-jax/README.md
```diff
@@ -0,0 +1,170 @@
+# Setup
```
ZYHowell, line 1: Please clean up this file for:
- the PyTorch part is not required;
- some code in this folder seems from [this repo](https://github.com/patil-suraj/stable-diffusion-jax), which uses the MIT license;
- provide with an example about how to configure parameters of the stable diffusion model in your example, i.e. how to configure the batch size, directory of models and outputs etc.

examples/stable-diffusion-jax/example.py
```diff
@@ -0,0 +1,89 @@
+import jax
```
ZYHowell, line 1: What's the difference among `alpa_run.py`, `example.py`, `nopmap.py` and `run.py`? If some are not related to alpa, could you please clean them up?

examples/stable-diffusion-jax/alpa_run.py
```diff
@@ -0,0 +1,145 @@
...
+    return pipe.sample(
+    input_ids,
+    uncond_input_ids,
+    prng_seed,
+    inference_state,
+    num_inference_steps,
+    guidance_scale,
+)
+
+inf_step = alpa.parallelize(inf_step, method=alpa.DataParallel())
```
ZYHowell, line 121: both the `input_ids` and `uncond_input_ids` are variables with a batch dimension, please set the [`batch_argnums`](https://github.com/alpa-projects/alpa/blob/bf62c1763c0fc3ec58e824ce421a010728af5ae5/alpa/api.py#L67) correctly.

ZYHowell (changes requested):
Thanks for contributing to Alpa! I left some comments about files in `examples/stable_diffusion_jax`. I only took a glimpse about the `examples/stable_diffusion_jax/stable_diffusion_jax` folder because it seems that they are directly copied from another repo, and I'll read files in the folder carefully once my comments are solved.
Plus, is there any way to install that repo instead of copying them? It's more convenient to install instead of copy because we can track their updates then.

Issue #752, opened 2022-10-18
[BUG] Strange pipeline parallel strategy when #device == #PP


PKUFlyingPig:
**Please describe the bug**
The produced executable is strange when manually setting the #pipeline parallel equals the available GPUs in the cluster.

**Please describe the expected behavior**
For example, with 4 p3.16xlarge instances, each has 8 GPUs, when set #pipeline parallel = 32, the output executable should contain 32 pipeline stages, where the 0-7th stage on the first node, 8-5th on the second node and so on, which can be notated as [[0-7],[8-15],[16-23],[24,31]]. But the actual output is: [[0, 23-29], [7-14], [15-22], [1-6, 30-31]]. As a result, the actual cross node communication increase from 3 to 5.

**System information and environment**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04, docker): Linux Ubuntu 18.04
- Python version: 3.9
- CUDA version: 11.1
- GPU model and memory: V100
- Alpa version: latest

**To Reproduce**
1. cd into `alpa/benchmark/alpa`
2. change the bottom three lines in [suite_inference_gpt.py](https://github.com/alpa-projects/alpa/blob/main/benchmark/alpa/suite_inference_gpt.py) into 
```
get_config(gpt_specs["2.6B"], [32], [1], [1], [1], [1])
```
3. run `python benchmark.py --suite gpt_inference.profile --profile-stage-execution-time`
4. download the generated chrome_tracing file in `./chrome_trace/bs=1,pp=32.json`
5.  open chrome with URL `chrome://tracing/` and load the json file, you can see the pipeline execution timeline.

**Screenshots**
![image](https://user-images.githubusercontent.com/60659728/196409851-97c5e933-6220-4b36-ae3f-e08644267920.png)

zhuohan123:
cc @ZYHowell

ZYHowell:
The problem comes from [this line](https://github.com/alpa-projects/alpa/blob/7894dc4a16b43760e8b54fb854b08d113ca791ea/alpa/pipeline_parallel/stage_construction.py#L647). If all mesh sizes are the same(`[1] * 32`), the order comes to be `0 29 28 27 26 25 24 23 22 21 20 19 18 17 16 15 14 13 12 11 10  9  8  7 6 5 4 3 2 1 30 31`. To reproduce the result, you can:
```python
np.argsort([1] * 32)
```
[...]

PR #757, opened 2022-10-20
[FIX] Use stable sort for all argsort in Alpa

zhuohan123:


zhuohan123:
Temporary fix of #752

PR #761, opened 2022-10-30
[BENCHMARK] Align the benchmark output format with the requirements of Alpa Serve

PKUFlyingPig:
- Add per-stage latency to benchmark output.
- To benchmark `bert-1.3b`, run `MODEL_NAME="bert-1.3b" python run_exp.py gpt_inference` under `benchmark/alpa/`, the output will be in `bert-1.3b.tsv` under the same folder.

### DIFF
--- a/alpa/device_mesh.py
+++ b/alpa/device_mesh.py
@@ -2033,6 +2033,14 @@ class PhysicalDeviceMeshGroup:
                 calls.append(worker.get_max_memory_allocated.remote())
         return max(ray.get(calls))
 
+    def get_max_memory_allocated_per_stage(self):
+        """Get the maximal size of memory allocated for each stage so far."""
+        return [mesh.get_max_memory_allocated() for mesh in self.meshes]
+
+    def reset_memory_stats(self):
+        for mesh in self.meshes:
+            mesh.reset_memory_stats()
+
     def destroy_collective_groups(self):
         for i in range(len(self)):
             for j in range(len(self)):
--- a/alpa/device_mesh.py
+++ b/alpa/device_mesh.py
@@ -2033,8 +2033,8 @@ class PhysicalDeviceMeshGroup:
                 calls.append(worker.get_max_memory_allocated.remote())
         return max(ray.get(calls))
 
-    def get_max_memory_allocated_per_stage(self):
-        """Get the maximal size of memory allocated for each stage so far."""
+    def get_max_memory_allocated_per_mesh(self):
+        """Get the maximal size of memory allocated for each mesh so far."""
         return [mesh.get_max_memory_allocated() for mesh in self.meshes]
 
     def reset_memory_stats(self):
--- a/benchmark/alpa/benchmark.py
+++ b/benchmark/alpa/benchmark.py
@@ -88,15 +88,16 @@ def benchmark_suite(suite_name,
         (parameter_count, peak_mem, latencies, tflops, metadata) = result
 
         heads = [
-            "ModelName", "BS", "#Microbatch", "DP", "OP", "PP", "#GPU", 
+            "ModelName", "BS", "#Microbatch", "DP", "OP", "PP", "#GPU",
             "MeanTime(s)", "StdTime(s)", "TFLOPs", "Weights(GB)", "PeakMem(GB)",
             "StageLatencies(s)"
         ]
         values = [
-            exp_name, benchmark_case.batch_size, num_micro_batches, parallel_args.dp, 
-            parallel_args.op, parallel_args.pp, num_gpus, f"{np.mean(latencies):.3f}",
-            f"{np.std(latencies):.3f}", f"{tflops:.2f}", f"{parameter_count/1e9:.3f}",
-            f"{peak_mem/GB:.3f}", metadata["avg_stage_latencies"]
+            exp_name, benchmark_case.batch_size, num_micro_batches,
+            parallel_args.dp, parallel_args.op, parallel_args.pp, num_gpus,
+            f"{np.mean(latencies):.3f}", f"{np.std(latencies):.3f}",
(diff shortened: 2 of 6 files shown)


benchmark/alpa/benchmark.py
```diff
@@ -88,16 +88,16 @@ def benchmark_suite(suite_name,
         (parameter_count, peak_mem, latencies, tflops, metadata) = result
 
         heads = [
-            "Type", "Model Config", "#Microbatch", "#GPU", "Parallel Config",
-            "Mean Time (s)", "Std Time (s)", "#Params (Billion)", "TFLOPs",
-            "Peak Mem (GB)", "Metadata"
+            "ModelName", "BS", "#Microbatch", "DP", "OP", "PP", "#GPU",
+            "MeanTime(s)", "StdTime(s)", "TFLOPs", "Weights(GB)", "PeakMem(GB)",
+            "StageLatencies(s)"
```
merrymercy, line 93: This script is shared by a lot of other benchmarks. Not all benchmarks have these fields. The default value `n_iters=10` is not good for other benchmarks.

We can either
1. Copy and port related things to mms repo or mms branch, so we don't need to worry about these compatibility issues in the future.
2. or add log in benchmark_one_case_gpt_bert_inference

merrymercy (changes requested):

benchmark/alpa/benchmark_one_case_gpt_bert_inference.py
```diff
@@ -170,16 +171,24 @@ def benchmark_gpt_inference_internal(model_type,
...
+        avg_stage_latencies = compute_avg_stage_latencies(timelines)
+        assert len(avg_stage_latencies) == num_manual_pipeline_stages
+    else:
+        avg_stage_latencies = None
 
     # Compute statistics
     tflops, parameter_count = compute_gpt_inference_statistics(
         benchmark_case, latencies, virtual_mesh.num_devices_per_host)
     metadata = {
         "latencies": latencies,
```
merrymercy, line 190: Why do we need this metadata? I think `latencies` is already in the return value

PKUFlyingPig: This `latencies` only contains end-to-end latency for each invocation.

PKUFlyingPig: This latencies only contains end-to-end latency for each invocation

merrymercy: Is it the same as the return value? https://github.com/alpa-projects/alpa/blob/c5ca9d63fab25956ec738f93cc48927cc25eab59/benchmark/alpa/benchmark_one_case_gpt_bert_inference.py#L211

merrymercy (approved):

benchmark/alpa/benchmark_one_case_gpt_bert_inference.py
merrymercy: Is it the same as the return value here?  https://github.com/alpa-projects/alpa/blob/c5ca9d63fab25956ec738f93cc48927cc25eab59/benchmark/alpa/benchmark_one_case_gpt_bert_inference.py#L211

The script already prints the mean and std of the latencies. Why are the mean and latency not enough?

merrymercy:
@PKUFlyingPig Please fix the conflicts

benchmark/alpa/benchmark_one_case_gpt_bert_inference.py
PKUFlyingPig: Because we need fine-grained per-stage latency?

PKUFlyingPig: No, the `latencies` is a list of end-to-end latency for each invocation. Now I removed the per-stage latency from metadata and only log it inside the benchmark_one_case_gpt_bert_inference if `profile_stage_execution_time` is set to True.

merrymercy: I meant `metadata` is necessary but L210 is not necessary. Because the variable `latency` in L210 is already in the return value.

alpa/device_mesh.py
```diff
@@ -2033,6 +2033,14 @@ def get_max_memory_allocated(self):
                 calls.append(worker.get_max_memory_allocated.remote())
         return max(ray.get(calls))
 
+    def get_max_memory_allocated_per_stage(self):
```
merrymercy, line 2036: per stage is not the same as per mesh

benchmark/alpa/run_exp.py
```diff
@@ -33,17 +35,20 @@ def run_exp(cluster_settings, suite_name, benchmark_settings=None):
...
+        "profile_stage_execution_time": True
+    }),
     "gpt_no_embedding_inference": ("gpt_no_embedding_inference.profile", {}),
     "gpt_inference_streaming": ("gpt_inference.profile", {
         "profile_driver_time": True
     }),
 }
-cluster_settings = [(8, 8), (4, 8), (2, 8), (1, 8), (1, 4), (1, 2), (1, 1)]
-
+cluster_settings = [(4, 8), (3, 8), (2, 8), (1, 8), (1, 4), (1, 2), (1, 1)]
```
merrymercy, line 47: Do not delete (8,8). There are other experiments using (8,8). 
I think (8,8) will be safely skipped for your case so adding it is okay.

benchmark/alpa/benchmark_one_case_gpt_bert_inference.py
PKUFlyingPig: I see. I'll remove it.

merrymercy (approved):

PR #764, opened 2022-11-03
[Feature] Support TPU backend for ShardParallel

ZYHowell:
This PR:
- Add `has_cuda` and `backend` in `global_config`
- Add `IS_CUDA` in `setup.py`
- Change the semantic of `run_auto_sharding_pass` and `run_backend_compilation`: the output of the `run_auto_sharding_pass` and input of `run_backend_compilation` is no longer the spmd-partitioned module, but instead the sharding annotated module. This is because TPU cannot compile an spmd-partitioned module with sharding configurations;
- write testcases specifically for TPUs. The pass that rewrites `AllReduce` to `ReduceScatter` and `AllGather` is after SPMD partitioner, so we cannot do it for the tpu backend.

### DIFF
--- a/alpa/__init__.py
+++ b/alpa/__init__.py
@@ -1,4 +1,12 @@
 """Alpa is a system for training large-scale neural networks."""
+from . import global_env
+
+from alpa.global_env import global_config
+try:
+    import cupy
+    global_config.has_cuda = True
+except ImportError:
+    global_config.has_cuda = False
 
 # Import all public packages
 from . import api
--- a/alpa/__init__.py
+++ b/alpa/__init__.py
@@ -1,7 +1,7 @@
 """Alpa is a system for training large-scale neural networks."""
 from . import global_env
 
-from alpa.global_env import global_config
+from .global_env import global_config
 try:
     import cupy
     global_config.has_cuda = True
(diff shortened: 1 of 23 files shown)


zhisbug:
One quick Q: how do we run TPU tests in CI?

ZYHowell:
> One quick Q: how do we run TPU tests in CI?

Google provides free TPU access for several days, but I've already extended the time twice(in the first extension they gave me 90 days and the second has 180 days). Can we use that machine for CI?

OhadRubin:
@merrymercy hey, it seems all the tpu specific tests are being skipped (on mobile, will provide a link soon).
So it is not clear to me what is the current working TPU capabilities.

ZYHowell:
> @merrymercy hey, it seems all the tpu specific tests are being skipped (on mobile, will provide a link soon). So it is not clear to me what is the current working TPU capabilities.

Currently we skip all tests other than those for ShardParallel/reduce-scatter optimization. You can see we rewrite only those tests, but not rewrite tests such as `test_n_layer_mlp_data_parallel`. For a full list of unit tests, please refer to the `if __name__ == "__main__":` part.

Issue #765, opened 2022-11-04
[RFC] Explicitly record HloStatus


ZYHowell:
**Describe the new feature and the current behavior/state**
Currently we implicitly assume `HloStatus` for the input and output of different functions in `auto_sharding.py`(but never check it) and have three different classes: `HloModule`, `XlaComputation` and serialized `bytes`, while using the same arg name `hlo_module`. This RFC will add a class `WrappedHlo`, which explicitly records the `HloStatus` and has the api to get any of the three classes above.

The reason to do so is the `backend_compilation` sometimes needs an input with status `SHARDING_ANNOTATED`(in TPU), but input with status `SPMD_PARTITIONED` should also be supported(we run `run_spmd_partition` for gradient accumulation to skip all-reduce in GPU backends).

**Will this change the current API? How?**
All `HloModule`, `XlaComputation` and `bytes` will be replaced by `WrappedHlo`s, unless it's a read-only arg.

ZYHowell:
https://github.com/alpa-projects/alpa/pull/766

PR #766, opened 2022-11-05
[Refactor] Use WrappedHlo to record HloStatus with HloModule

ZYHowell:
See https://github.com/alpa-projects/alpa/issues/765


ZYHowell:
cc @zhuohan123. This PR changes some internal api in stage profiling, so https://github.com/alpa-projects/alpa/pull/681 may change accordingly.

PR #785, opened 2022-11-26
[BENCHMARK] Add scripts to generate profiling database for alpa serve

PKUFlyingPig:
Add a single script to generate the profiling database for alpa_serve automatically.


benchmark/alpa/gen_benchmark_database.py
```diff
@@ -0,0 +1,13 @@
+import argparse
```
merrymercy, line 1: rename this to `gen_serving_database.py`

merrymercy (approved):

Issue #776, opened 2022-11-14
Encounter a 'profile worker forcely killed' & NotImplementedError (auto-slicing layers with existing physical meshes) when trying to enlarge the global batch size.


DicardoX:
Hi, there. I'm trying to run Alpa PipeShardParallel training for a ResNet50 model (CIFAR10 dataset) on a Distributed Ray Cluster with two 1080ti (4-GPU each) servers. 

- The **parallel method** is as follows:

```python
method = alpa.PipeshardParallel(num_micro_batches=self.num_micro_batches,                                   										layer_option=alpa.AutoLayerOption(layer_num=self.pipeline_layer_num),
                                        stage_option="auto")
```

- The **training configs** are as follows:

```python
trainer_cfgs = {
        'model_name': 'wide_resnet',
        'dataset_name': 'CIFAR10',
        'batch_size': 16,
        'lr': 1e-3,
        'momentum': 0.9,
        'rand_seed': 123,
        'dtype': jnp.float32,
        'num_micro_batches': 16,
        'pipeline_layer_num': 16,
        'parallel_mode': 'search',
        'niter': 12,
        'profile_driver_time': True,
    }
```

- The **topology of the Ray cluster** is:

1080ti-01 (head node, 4 GPUs) -> 1080ti-02 (worker node, 4 GPUs).

- The **Alpa cluster is initialized** as:

```python
ray.init(address="auto")
alpa.init(cluster="ray", num_devices_per_node=2, num_nodes=2)
```

As shown above, when I set the batch size to 16, everything went well and I can finish the training process. However, when I **enlarged the batch size to 64**, the following errors occurred:

- **When profiling submesh (1, 1)** (the above profile for (2, 2) and (1, 2) went well):

    ```bash
    - Profiling for submesh 0 (1, 1):
    - Generate all stage infos (Jaxpr -> HLO)
    - Compile all stages
    - Profile all stages
    WARNING:alpa.pipeline_parallel.stage_profiling:Meet unexpected error, all profile workers are forcely killed
[...]

PR #796, opened 2022-12-02
bypass nodes which don't have gpu resources.

dumpmemory:
try to pass the nodes which don't have gpu resources.

in my setting, the ray head node don't have any cpu or gpu resources so we might need to skip these nodes.


dumpmemory:
i might need to consider this  #https://github.com/alpa-projects/alpa/issues/776#issue-1447588779

merrymercy:
This seems to be a reasonable fix.

1. Could you fix the lint error?
2. Did this solve your problem?  I guess there might be other places we also need to bypass nodes without GPU resources.

dumpmemory:
1. I will try to fix the lint error
2. no i still facing the problem like #https://github.com/alpa-projects/alpa/issues/776#issue-1447588779. the grpc is failed. in my ray cluster setting, the training process is deployed on the worker node with gpus, which means there is no workload on head node. So there must be other places need to modified 
3.

dumpmemory:
1.  the grpc issues should be fixed by https://github.com/alpa-projects/alpa/pull/803#issue-1478530799

ZYHowell:
@dumpmemory sorry I missed point 3 in your last reply and thought the problem in this issue is fully addressed , so I merged main into this branch with some change as a pre-merge process. If you have other commits, please forcely push to cover mine

dumpmemory:
> @dumpmemory sorry I missed point 3 in your last reply and thought the problem in this issue is fully addressed , so I merged main into this branch with some change as a pre-merge process. If you have other commits, please forcely push to cover mine

For the third question is that  i can't use PipeshardParallel to find a parallel setting for 1.3b gpt-neo model on 8 2080 ti gpus. This might be because i am not  familiar with the alpa yet. which might not related to this pr. pls ignore it. Thanks again

dumpmemory:
1. I have test it on my side by running https://github.com/alpa-projects/alpa/blob/main/tests/pipeline_parallel/test_bert.py and https://github.com/alpa-projects/alpa/blob/main/examples/gpt2/run_clm_flax.py （with gpt-neo 125m ） and it all passed. 
2. I haven't run https://github.com/alpa-projects/alpa/blob/main/tests/shard_parallel/test_bert.py. But when i modify the run_clm_flax.py using ShardParallel method, it failed with #810. it seems that the peak_memory is inf.

dumpmemory:
The bypass GPU pr pass https://github.com/alpa-projects/alpa/blob/main/tests/shard_parallel/test_bert.py test.

zhuohan123 (approved):
LGTM

Issue #826, opened 2022-12-21
jaxlib.xla_extension.XlaRuntimeError: DEADLINE_EXCEEDED: Connect() timed out after 0 with 1 attempts. 


dumpmemory:
**Please describe the bug**

I can't running https://github.com/dumpmemory/alpa/blob/49818ad85da01a52f86e3c2b92e9681d991c4639/examples/opt_finetune/test_run_clm_flax_opt.py with gpt-neo 1.3b, i have removed deterministic args. using following command

ray version is 2.2.0
```
 ray job submit --working-dir . --address http://10.32.10.198:8265/ --  python3 run_clm_flax_opt.py     --output_dir="/mnt/train/test_gpt-neo-1.3b"     --model_name_or_path="EleutherAI/gpt-neo-1.3B"     --dataset_name="wikitext"     --dataset_config_name="wikitext-2-raw-v1"     --do_train --do_eval     --block_size="1024"     --per_device_train_batch_size="2"     --per_device_eval_batch_size="2"     --num_micro_batches 2     --operator_parallel 2     --pipeline_parallel 4     --dtype="float16"     --learning_rate="5e-4" --warmup_steps="2000"     --adam_beta1="0.9" --adam_beta2="0.98" --weight_decay="0.01"     --overwrite_output_dir     --num_train_epochs="8"     --logging_steps="16"     --save_steps="2500"     --eval_steps="2500"
```

**Please describe the expected behavior**

System information and environment
Python version: 3.8
CUDA version: 11.3
NCCL version: 820
cupy version: cupy-cuda113
GPU model and memory: 2080ti 11g
Alpa version: install from source
TensorFlow version: 2.1.0
JAX version: 0.3.22+cuda113.cudnn820

**To Reproduce**
1.
2.
3.
Error log
```
(main pid=37087, ip=10.32.13.122) INFO:__main__:***** Running training *****
(main pid=37087, ip=10.32.13.122) INFO:__main__:  Num examples = 2318
(main pid=37087, ip=10.32.13.122) INFO:__main__:  Num Epochs = 8
(main pid=37087, ip=10.32.13.122) INFO:__main__:  Batch size per device (w. accumulation) = 2
(main pid=37087, ip=10.32.13.122) INFO:__main__:  Global train batch size (w. parallel & distributed) = 16
(main pid=37087, ip=10.32.13.122) INFO:__main__:  Total optimization steps = 1152
(main pid=37087, ip=10.32.13.122)
[...]

Issue #797, opened 2022-12-02
Key Error for  buffer_dict


dumpmemory:
**Please describe the bug**

Then I tried to train a gpt-neo 1.3b  parameters model using the code https://github.com/alpa-projects/alpa/blob/main/examples/gpt2/run_clm_flax.py on 2 nodes, one head node which don't have any cpus or gpus and one worker node has 8 gpus showing the following error:
~~~
2022-12-02 14:31:03,423	ERROR worker.py:400 -- Unhandled error (suppress with 'RAY_IGNORE_UNHANDLED_ERRORS=1'): ray::MeshHostWorker.run_executable() (pid=8620, ip=10.32.10.64, repr=<alpa.device_mesh.MeshHostWorker object at 0x7fd3c60e67c0>)
  File "/home/nlp/miniconda/lib/python3.8/site-packages/alpa/device_mesh.py", line 277, in run_executable
    self.executables[uuid].execute_on_worker(*args, **kwargs)
  File "/home/nlp/miniconda/lib/python3.8/site-packages/alpa/mesh_executable.py", line 907, in execute_on_worker
    del buffer_dict[next_batches_uuids[i]]
KeyError: None
~~~

**Please describe the expected behavior**

**System information and environment**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04, docker): Docker based on nvidia/cuda:11.3.1-cudnn8-devel-centos7
- Python version: 3.8
- CUDA version: 11.3
- NCCL version: 820
- cupy version: cupy-cuda113
- GPU model and memory: 2080ti 11g
- Alpa version: 0.2.2
- TensorFlow version: 2.1.0
- JAX version: `0.3.22+cuda113.cudnn820`

**To Reproduce**

I have modified the https://github.com/alpa-projects/alpa/blob/75f3597dbf4a49debce3cc1a8c83f2cb7b8b6eda/alpa/device_mesh.py#L2153 line, to pass the node which don't have any gpus. 

Then I tried to train a gpt-neo 1.3b parameters model using the code https://github.com/alpa-projects/alpa/blob/main/examples/gpt2/run_clm_flax.py on 2 nodes, one head node which don't have any cpus or gpus and one worker node has 8 gpus(2080) using following command :
~~~
[...]

PR #830, opened 2022-12-22
Add xla server port config and pass next batches uuids ==(none,)

dumpmemory:
1. add xla server port config when the docker port are limited to special range  like 8000~8999, should fix  https://github.com/alpa-projects/alpa/issues/826 
2. conner case when next_batches_uuids ==(None,) should fix https://github.com/alpa-projects/alpa/issues/797
3.

### DIFF
--- a/alpa/device_mesh.py
+++ b/alpa/device_mesh.py
@@ -26,6 +26,7 @@ from operator import attrgetter
 import os
 import pickle
 import shutil
+import socket
 import threading
 import time
 from typing import Any, List, Union, Sequence, Tuple, Optional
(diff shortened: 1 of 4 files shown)


dumpmemory:
The first issue is not totally fixed.

dumpmemory:
Another is when the run_clm_flax_opt.py job finished, the port was not closed as expect

merrymercy:
Thanks for fixing these bugs! It is merged.

dumpmemory:
> Thanks for fixing these bugs! It is merged.

For the first issues, the used port is from ray side. When the job finished, the port was still used by IDLE which was from main function. i am checking it. Thanks .

dumpmemory:
fixed, when  adding ray.remote to the main function in https://github.com/dumpmemory/alpa/blob/49818ad85da01a52f86e3c2b92e9681d991c4639/examples/opt_finetune/test_run_clm_flax_opt.py , must add alpa.shutdown() to close the port.

PR #840, opened 2023-01-01
Add k8s/infiniband build and deployment guide to Alpa

jiaodong:
Add complete guide to deploy Alpa on CoreWeave cloud (k8s with infiniband). See added documentation page for more details.

In addition, this PR makes a few changes to current Alpa codebase:
- Changed `profile_communication.py` to add an `ib` flag
- Changed docker documentation page to use `main` branch instead of `master`
- Changed default wheel from `jaxlib==0.3.15` to `jaxlib==0.3.22`.
- Added keys to `docker/build_jaxlib.Dockerfile`

### DIFF
--- a/benchmark/cupy/profile_communication.py
+++ b/benchmark/cupy/profile_communication.py
@@ -228,6 +228,8 @@ if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument("--efa", action="store_true",
         help="Use AWS EFS on p3.24 or p4.24 instances")
+    parser.add_argument("--ib", action="store_true",
+        help="Use InfiniBand for NCCL communcation")
     parser.add_argument("--debug", action="store_true",
         help="Print nccl debug information")
     args = parser.parse_args()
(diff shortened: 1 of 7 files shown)


benchmark/cupy/profile_communication.py
```diff
@@ -245,6 +247,13 @@ def sync(self):
                 "FI_EFA_USE_DEVICE_RDMA": "1",
                 "LD_LIBRARY_PATH": os.environ.get("LD_LIBRARY_PATH", ""),  # For libnccl-net.so
             }
+        elif args.ib:
+            env_vars = {
+                "NCCL_SOCKET_NTHREADS": "4",
+                "NCCL_NSOCKS_PERTHREAD": "4",
+                "NCCL_IB_HCA": "ibp",
```
zhisbug, line 254: this is hard code? because other machines might have different IB interface, e.g., `mlx5`

zhisbug (approved):

benchmark/cupy/profile_communication.py
jiaodong: that's right, for CoreWeave mlx5 is for ethernet and ibp is for inifiniband, but i remember for nvidia's case it could be different depending on host type

PR #845, opened 2023-01-04
[Docs]Update jaxlib build from source

jiaodong:
Most AWS EC2 docker images come with gcc 9.4 but will run into issues if trying to build Alpa from source. Updating to gcc 10 with clean bazel cache fixes it, thus updating developer guide for the community.


merrymercy:
I think these instructions are already in  https://github.com/alpa-projects/alpa/blob/d1d9aa8fab31bec6f51d063b731cc0159009e73e/docs/install.rst#method-2-install-from-source

jiaodong:
i see, this is indeed duplicated info then. I think my confusion mostly come from structure of the doc since developer guide part didn't mention the info needed to build & iterate tensorflow-alpa that is included in `install` section. How about I change this PR to link developer guide to installation ?

merrymercy:
I don't think we need this because "Install" is the default place to find instructions for building the project.

Issue #700, opened 2022-09-13
[FEATURE] OPT-175B service authentication and new priority queue


zhisbug:
## Background

Currently, the OPT serving backend does not have a good priority mechanism. It is (mostly) a first-come-first-serve queue.
The only priority rule is that the `completion` endpoint has a higher priority than the `logprob` endpoint.

We would like to implement the following new priority order (from high to low) in the [batching loop](https://github.com/alpa-projects/alpa/blob/5cd7f7dd6eddf687bff895543883c6a0388e8451/examples/llm_serving/launch_model_worker.py#L82).
  - 1st (interactive jobs): Requests from website users who pass the recaptcha test. These are low-frequent jobs.
  - 2nd (background jobs): Requests from client library users with API keys. To share a single backend among all users, we can use weighted round-robin or lottery scheduling. We can assign different weights to different API keys.
  - 3rd: Requests from users without API keys

With the new priority mechanism, we want to only assign priority according to the source of the requests and deprecate the old rule. `completion` endpoint and `logprob` endpoint will have the same priority.

## Useful pointers
- The existing priority queue https://github.com/alpa-projects/alpa/blob/5cd7f7dd6eddf687bff895543883c6a0388e8451/examples/llm_serving/launch_model_worker.py#L54
- How to distinguish interactive requests from the website and requests from the client library. Interactive requests from the website have Recaptcha information attached. Background requests from the client library have API keys attached.  https://github.com/alpa-projects/alpa/blob/5cd7f7dd6eddf687bff895543883c6a0388e8451/examples/llm_serving/launch_model_worker.py#L321-L331

zhisbug:
@aurickq

JubilantJerry:
@Weltberg and I are thinking about contributing this feature. There's several possibilities for how the weighting could be done, I wonder what you think would be the best for this project?

1. We could strictly pick high priority tasks over lower priority tasks all the time. As I understand it, the current code always prioritizes `completion` tasks whenever they exist, only running `logprob` tasks if no `completion` tasks are queued. Something similar could be done for the 3 priority classes introduced in the feature. But you mention weighted round robin and lottery scheduling, which still runs the lower priority tasks sometimes. I imagine this is a good thing since it avoids starvation, so I'm pretty sure you don't want strict priority scheduling. But I just wanted to clarify.

[...]

PR #848, opened 2023-01-07
add WRR priority algo to launch model worker

FedericoCampe8:


merrymercy (changes requested):
The design looks good. According to #700, there is still one thing we want: assigning different priority levels to different API keys. We want to load the priorities of API keys from the key file.
https://github.com/alpa-projects/alpa/blob/5b1bb8d5a9f1c3e2220154ec4d2bc789dcca3e9b/examples/llm_serving/launch_model_worker.py#L132-L133

Let us use the following default priority:
The default priority of interactive requests is 100.
The default priority of an API key is 10.
The custom priority of an API key can be specified in `KEY_FILE`

examples/llm_serving/launch_model_worker.py
```diff
@@ -79,22 +132,66 @@ def __init__(self,
...
+                request_queue = self.get_priority_queue()
+
+            # The nowait should not throw since there is at least one item
+            # in the current request queue
+            pri_item = request_queue.get_nowait()
+            if pri_item is None:
+                # The current queue can be empty,
+                # e.g., for interactive jobs that are low-frequency.
+                # In this case continue until there is a queue with an item
+                continue
```
merrymercy, line 178: I guess it cannot be empty because you have a while loop above and the request_queue is always not empty.
Am I correct?

examples/llm_serving/launch_model_worker.py
```diff
@@ -22,12 +23,47 @@
...
+
+    def get_weight(self):
+        return self.weight
 
 
 GenerateItem = namedtuple("GenerateItem", ["uid", "return_queue", "data"])
 LogprobsItem = namedtuple("LogprobsItem", ["uid", "return_queue", "data"])
 
+GENERATE_ITEM_PRIORITY = 0
+LOGPROBS_ITEM_PRIORITY = 50
```
merrymercy, line 58: This priority can be deprecated now. Let us just use a single-level priority queue instead of the current nested two-level priority queue.

i.e.
Put all requests into a single priority queue. Only set the priority according to their source (i.e. interactive or background). Do not set priorities according to their request endpoint (i.e. generate or logprobs)

examples/llm_serving/launch_model_worker.py
FedericoCampe8: Yes, you are. This is a leftover, will remove.

examples/llm_serving/launch_model_worker.py
FedericoCampe8: sounds good. I thought that having two or more queues would provide more flexibility for the future but sure, we can have everything in one queue. Will do.

FedericoCampe8: That would default to what it is currently on main with the exception of changing the old priorities based on the source, wouldn't it? Unless I understand it wrong, we still need 2 queues - one for each type of source to avoid starvation. In other words, I need to rotate between interactive and background otherwise background (low priority) can starve if new interactive jobs come in. If instead, I pick 1 background for every n interactive (and background itself has priority given the key), then it would avoid starvation. Am I right or am I missing something?

merrymercy: You are right. In the old code, `generate_item` always has a higher priority than `logprob_item`. Now we want to remove this design. `generate_item` and `logprob_item` will have the same priority.
We only assign priority according to sources (iterative vs background). In other words,
Old rule:
logprob_item > generate_item
New rule:
interactive logprob_item = interactive generate_item > background logprob_item = background generate_item

merrymercy:
Deprecated in favor of #852

PR #909, opened 2023-03-30
Use shell script to drive jaxlib build

yhtang:
Together with https://github.com/google/jax/pull/15205, this should eliminate the need to apply the `build_jaxlib/update_build_scripts.patch` each time we rebase. Note the `--editable` option will only be usable once alpa is rebased onto JAX 0.4.8.

The original `update_build_scripts.patch` file is meant to provide two functionalities:

1. build jaxlib as an 'editable' package. This can now be done by
```
cd build
./build.sh  # build wheel
./build.sh --editable  # build editable installation
```

2. Ask Bazel to prefer local installations NCCL. This can now be done by
```
./build.sh --nccl local  # prefer local NCCL
./build.sh --nccl download  # prefer to use XLA's pinned version
```


build_jaxlib/build.sh
```diff
@@ -0,0 +1,287 @@
...
+    elif [[ "$CUDA_COMPUTE_CAPABILITIES" == "local" ]]; then
+        export TF_CUDA_COMPUTE_CAPABILITIES=$(./local_cuda_arch)
+    else
+        export TF_CUDA_COMPUTE_CAPABILITIES="${CUDA_COMPUTE_CAPABILITIES}"
+    fi
+fi
+
+if [[ "${EDITABLE}" == "1" ]]; then
+    BUILD_PARAM="${BUILD_PARAM} --editable"
+fi
```
gjoliver, line 190: why don't we do this directly in the flag parsing while loop above?
you then don't need the additional EDITABLE variable.

build_jaxlib/build.sh
```diff
@@ -0,0 +1,287 @@
...
+
+if [[ -d /cache ]]; then
+    BUILD_PARAM="${BUILD_PARAM} --bazel_options=--disk_cache=/cache"
+fi
+
+if [[ "$DEBUG" == "1" ]]; then
+    BUILD_PARAM="${BUILD_PARAM} --bazel_options=-c --bazel_options=dbg --bazel_options=--strip=never --bazel_options=--cxxopt=-g --bazel_options=--cxxopt=-O0"
+fi
```
gjoliver, line 198: same for --debug I guess ...

build_jaxlib/build.sh
```diff
@@ -0,0 +1,287 @@
...
+print_var SRC_PATH_XLA
+
+print_var TF_CUDA_VERSION
+print_var TF_CUDA_COMPUTE_CAPABILITIES
+print_var TF_CUBLAS_VERSION
+print_var TF_CUDNN_VERSION
+print_var TF_NCCL_VERSION
+print_var CC_OPT_FLAGS
+
+print_var BUILD_PARAM
```
gjoliver, line 224: BUILD_PARAM is printed twice. get rid of the one above I guess?

build_jaxlib/local_cuda_arch
```diff
@@ -0,0 +1,30 @@
+//usr/bin/env nvcc -odir `mktemp -d` -x cu --run "$0" ${1:+--run-args "${@:1}"} ; exit $?
```
gjoliver, line 1: is this a new util?

gjoliver (commented):
nice cleanup.
let's figure out when we should merge this.

PR #914, opened 2023-04-04
Moved _compute_one_replica_ids from DistributedArray to PhysicalDeviceMesh

### LINKED ISSUE #837: [PERF] Move cached property out of `DistributedArray`
Background
---
To manage a tensor distributed on a device mesh, we use `DistributedArray` to represent this concept on the driver side. The class implements apis to load, save, and retrieve data from devices to the driver. Because the tensor is sharded on the device mesh, the `DistributedArray` should know how to get all parts of the data, see [this property](https://github.com/alpa-projects/alpa/blob/main/alpa/device_mesh.py#L1638-L1643). The `DistributedArray` is read-only, so we cache the property.
However, we found `DistributedArray` exceptionally light. In each iteration(a.k.a. generating a token), the driver needs to get a tiny tensor from devices. Although the tensor's sharding spec is the same for each iteration, the `DistributedArray` object differs, and the property needs to be repeatedly computed for each iter.
In this issue, we want to move computing `one_replica_buffer_ids` from `DistributedArray` to `PhysicalDeviceMesh` or even just a util function, because once the mesh shape and sharding spec are kept, we can know how to get and assemble shards to cover the whole tensor.

TODO
---
- [ ] Move `compute_one_replica_ids` to `PhysicalDeviceMesh` or `util`. It only uses sharding spec and mesh shape as inputs, regardless of which array it computes;
- [ ] Support `ReplicatedDistributedArray`.

akhmedsakip:
closes #837 (1st part)

Although the changes do not significantly reduce total execution time for use cases like _llm_serving (e.g., examples/llm_serving -> python3 textgen.py --model alpa/opt-2.7b)_, new dictionary lookups are generally faster by a factor of 5-10 than computing one_replica_ids.

### DIFF
--- a/alpa/device_mesh.py
+++ b/alpa/device_mesh.py
@@ -638,6 +638,7 @@ class PhysicalDeviceMesh(ABC):
     num_devices_per_host: int
     mesh_id: int
     operation_executables: dict
+    one_replica_ids: dict
 
     def get_signature(self) -> str:
         """Return a signature string that contains the mesh shape and GPU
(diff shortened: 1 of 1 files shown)


ZYHowell (changes requested):
lgtm. Some minor requests:
1. can you please run the `format.sh` file to reformat the code?
2. The `_one_replica_buffer_ids` and `_one_replica_host_local_ids` are no longer used, please remove them;
3. the `TODO` line above `one_replica_buffer_ids` can be removed now.

ZYHowell (approved):

PR #915, opened 2023-04-04
Allow users to specify a different cluster address when initializing Ray

### LINKED ISSUE #913: Specify Ray instance port
**System information**
- Alpa version: 0.2.3
- Are you willing to contribute it (Yes/No): Probably not qualified

**Describe the new feature and the current behavior/state**
Currently, Apla is fixed to connecting to port 6379 of a Ray instance. This limits the number of users in a node with multiple GPUs to 1 (e.g., if there are 4 GPUs, 2 users cannot use 2 GPUs each). While it is possible to start a Ray instance listening on a different port, Alpa seems to use 6379 by default, but no 2 Ray instances can listen on the same port. Would it be possible to have this port specified in a configuration file or in the arguments, so that a 2nd user could start a Ray instance on a port different than 6379?

gjoliver:
Closes #913

### DIFF
--- a/alpa/api.py
+++ b/alpa/api.py
@@ -23,6 +23,7 @@ is_initialized = False
 
 
 def init(cluster: str = "ray",
+         cluster_address: Optional[str] = None,
          num_nodes: Optional[int] = None,
          num_devices_per_node: Optional[int] = None,
          namespace: Optional[str] = "alpa_default_space"):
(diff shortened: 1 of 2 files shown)


merrymercy (approved):

ZYHowell (approved):

gjoliver:
Thanks

PR #921, opened 2023-04-20
removed mapping of numpy.int and numpy.float to nccl and gloo types

sammeralomair:
removed from mapping because they're deprecated


gjoliver:
which version were these removed from?
should we add it as package requirement?

sammeralomair:
1) v0.2.3
2) if using numpy 1.24 installing alpa will give an error so required

ZYHowell (approved):

PR #938, opened 2023-06-18
task 801 - launch physical meshes after compilation

haifaksh:


alpa/device_mesh.py
```diff
@@ -2305,6 +2305,78 @@ def profile_all(self, *args, **kwargs):
         return mesh_profiling.profile_all(self, *args, **kwargs)
 
 
+#TODO Github Task - CustomVirtualMesh for interfaces
+class VirtualWorker:
+    def __init__(self, index):
+        self.index = index
+        # Additional attributes or methods of virtual workers
+
+class CustomVirtualMesh(VirtualPhysicalMesh):
```
ZYHowell, line 2314: The name "Custom" seems weird. I'd suggest words like "Dummy" to represent that this is just a placeholder.

alpa/device_mesh.py
```diff
@@ -2305,6 +2305,78 @@ def profile_all(self, *args, **kwargs):
...
+    def __init__(self,
+                 host_ids: Sequence[int],
+                 host_info: Sequence[dict],
+                 num_devices_per_host,
+                 parent: "VirtualPhysicalMesh" = None,
+                 devices: Sequence[Sequence[int]] = None,
+                 mesh_id: int = None
+                 ):
+        super().__init__(host_ids, host_info, num_devices_per_host, parent, devices)
+        self.host_ips = []
```
ZYHowell, line 2324: This member seems never used. If so, please remove it

alpa/pipeline_parallel/compile_executable.py
```diff
@@ -245,8 +251,16 @@ def compile_pipeshard_executable_internal(
     debug_compilation_time("shard stages")
 
     # Launch the physical mesh group
-    if virtual_mesh.launched_physical_mesh_group is None:
-        virtual_mesh.get_physical_mesh_group(sliced_virtual_meshes)
+    # if virtual_mesh.launched_physical_mesh_group is None:
+    #     virtual_mesh.get_physical_mesh_group(sliced_virtual_meshes)
+
+    nccl_instantiated = False
+    if 'virtual_meshes' in globals() and virtual_meshes is not None and virtual_mesh.launched_physical_mesh_group is not None:
```
ZYHowell, line 258: please avoid using a global `virtual_meshes`. Instead, please:
1. Always create a virtual mesh group without launching NCCL. Remember we always don't want to touch any physical resources related information in this part.
2. Launch meshes and nccl groups on real resources in the `PipeShardDriverExecutable`'s init function. where the virtual mesh group checks if there is already a physical mesh group. If not, it creates a new one. Otherwise, it checks whether it's compatible with the existing group and throws an error when it's not compatible

alpa/pipeline_parallel/compile_executable.py
```diff
@@ -274,7 +289,8 @@ def compile_pipeshard_executable_internal(
     pipeshard_config = emitter_cls(**emitter_kwargs).compile()
 
     debug_compilation_time("runtime emitter")
-    return pipeshard_config
+    return pipeshard_config, sliced_virtual_meshes, virtual_meshes
```
ZYHowell, line 292: Move return values you add into the `pipeshard_config` by creating new members or changing the semantic of the existing ones

alpa/pipeline_parallel/cross_mesh_resharding.py
```diff
@@ -195,6 +195,8 @@ def __init__(self, task_spec, collective_group, src_mesh, dst_mesh):
         self.send_worker_task_ids = {}
         self.recv_worker_task_ids = {}
 
+        self.task_dones = []
```
ZYHowell, line 198: remove unused code. For a series of tasks, we always create a tmp task dones instead of keeping using the same one

alpa/pipeline_parallel/pipeshard_executable.py
```diff
@@ -63,7 +82,20 @@ def __init__(self,
         self.flop_count = pipeshard_config.flop_count
         self.stage_input_shard_specs = pipeshard_config.stage_input_shard_specs
         self.input_placement_specs = pipeshard_config.input_placement_specs
+        #TODO Github Task - Adding these lines
+        # self.env = pipeshard_config.env
```
ZYHowell, line 86: remove unused lines.

alpa/pipeline_parallel/pipeshard_executable.py
```diff
@@ -90,12 +122,83 @@ def __init__(self,
...
+        temp_worker_to_rank_map = {
+            worker: r for r, worker in enumerate(temp_mesh_grp)
+        }
+        for cgp in self.mesh_group.collective_groups:
+            for cg in cgp:
+                if cg is not None:
+                    cg.mesh_workers = temp_mesh_grp
+                    cg.worker_to_rank_map = temp_worker_to_rank_map
+        for virtual_worker, _ in pipeshard_config.instruction_lists.items():
+            virtual_to_pysical_map[virtual_worker.index] = virtual_worker
```
ZYHowell, line 149: Move codes to launch on real physical resources into a method of `VirtualMeshGroup`(probably name it `instantiate`). The target of the function is to lower a `VirtualMeshGroup` into a `PhysicalDeviceMeshGroup`. The pseudo-code of the method is:
```python
def instantiate(self, *args):
    if exist_a_physical_mesh_group():
        existing_group = get_existing_physical_mesh_group()
        self._check_compatibility(existing_group)
        return existing group
    else:
        return self._instantiate_mesh_group()
```

alpa/pipeline_parallel/pipeshard_executable.py
```diff
@@ -90,12 +122,83 @@ def __init__(self,
...
+                    uuid = resharding_task.broadcast_worker_task_ids[v_worker]
+                    worker = resharding_task.collective_group.mesh_workers[v_worker.index]
+                    task_dones.append(
+                    worker.put_resharding_broadcast_task.remote(
+                            uuid, task, resharding_task.collective_group.group_name))
+                ray.get(task_dones)
+
+        ##### For cross-mesh resharding #####
+        if not self.v_meshes.launched_nccl:
+            self._instantiate_nccl_groups(pipeshard_config.device_str_groups)
```
ZYHowell, line 200: as mentioned above, we always instantiate or inherit from an existing physical mesh group before this step, so we just use the original code.

ZYHowell (changes requested):

tests/runtime/test_create_state.py
```diff
@@ -100,8 +100,8 @@ def test_pipeshard_parallel(self):
 def suite():
     suite = unittest.TestSuite()
     suite.addTest(CreateStateTest("test_shard_parallel"))
-    suite.addTest(CreateStateTest("test_shard_parallel_grad_acc"))
-    suite.addTest(CreateStateTest("test_pipeshard_parallel"))
+    #suite.addTest(CreateStateTest("test_shard_parallel_grad_acc"))
```
ZYHowell, line 103: recover the unit test

